Issue #47, opened 2021-04-08
Add more examples


seanmor5:
Willing to accept examples on different datasets and models to demonstrate different parts of the Axon API and to demonstrate Axon's viability in the ecosystem. The TensorFlow guides are a great place to look for different datasets and problems. If you're blocked on any specific issue feel free to comment on the relevant issue with your use case :)

tiagodavi:
Hi @seanmor5 I am trying to create an example to predict diabetes with Axon and Nx, but I am still trying to understand how it works. 

Currently I have an error:

X = #Nx.Tensor<
  f32[615][8]
  
Y =  #Nx.Tensor<
  s64[615]
 
 This is the code I'm trying to create: 
 https://gist.github.com/tiagodavi/a905abeaf4d1f92c21f9df9043d196fe
 
```
StreamExecutor device (0): Host, Default Version
** (ArgumentError) expected input shapes to be equal, got {615} != {615, 1}
    (axon 0.1.0-dev) lib/axon/shared.ex:22: anonymous fn/1 in Axon.Shared."__defn:assert_shape!__"/2
    (nx 0.1.0) lib/nx/defn/compiler.ex:114: Nx.Defn.Compiler.__remote__/4
    (axon 0.1.0-dev) lib/axon/losses.ex:122: Axon.Losses."__defn:binary_cross_entropy__"/3
    (axon 0.1.0-dev) lib/axon/loop.ex:325: anonymous fn/5 in Axon.Loop.train_step/3
    (nx 0.1.0) lib/nx/defn/grad.ex:20: Nx.Defn.Grad.transform/3
    (axon 0.1.0-dev) lib/axon/loop.ex:332: anonymous fn/4 in Axon.Loop.train_step/3
    (axon 0.1.0-dev) lib/axon/loop.ex:1135: anonymous fn/4 in Axon.Loop.build_batch_fn/2
    (nx 0.1.0) lib/nx/defn/compiler.ex:101: Nx.Defn.Compiler.runtime_fun/4
```

seanmor5:
Hi @tiagodavi! Axon's implementation of BCE expects `y_true` to have a last dimension of size 1 (there's an explicit check for shape equality between `y_true` and `y_pred`). If you add a new axis to your `y_true`: `Nx.new_axis(y, -1)` - then the error should go away.

We can probably relax the strictly equal shape constraint, feel free to open a PR otherwise I will open an issue to track.

[...]

PR #48, opened 2021-04-08
FashionMNIST dataset and basic autoencoder

t-rutten:
Related to #47.

### DIFF
--- a/examples/fashionmnist_autoencoder.exs
+++ b/examples/fashionmnist_autoencoder.exs
@@ -42,14 +42,14 @@ end
 defmodule Autoencoder do
   def encoder(x) do
     x
-    |> Axon.dense(64, activation: :relu)
-    |> Axon.dense(3, activation: :relu)
+    |> Axon.dense(64, activation: :tanh)
+    |> Axon.dense(3, activation: :tanh)
   end
 
   def decoder(x) do
     x
-    |> Axon.dense(64, activation: :relu)
-    |> Axon.dense(784, activation: :relu)
+    |> Axon.dense(64, activation: :tanh)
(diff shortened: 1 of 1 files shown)


examples/fashionmnist_autoencoder.exs
```diff
@@ -0,0 +1,85 @@
...
+train_images
+|> Enum.at(0)
+|> (&(&1[:rand.uniform(32)])).()
+|> Nx.reshape({28, 28})
+|> Nx.to_heatmap()
+|> IO.inspect()
+
+IO.puts("\nTraining autoencoder...")
+
+final_params =
```
seanmor5, line 80: This variable would actually end up being:

```elixir
{final_params, _optimizer_state} =
```

I need to change the other examples as well

seanmor5 (commented):
@t-rutten This is great, I've been wanting to write an Autoencoder for awhile :)

Just one thing, can you add a prediction and display it as a heatmap at the end?

examples/fashionmnist_autoencoder.exs
```diff
@@ -0,0 +1,85 @@
...
+  def encoder(x) do
+    x
+    |> Axon.dense(64, activation: :relu)
+    |> Axon.dense(3, activation: :relu)
+  end
+
+  def decoder(x) do
+    x
+    |> Axon.dense(64, activation: :relu)
+    |> Axon.dense(784, activation: :relu)
```
seanmor5, line 52: Might it be better if this is `:sigmoid` or `:tanh` since the image inputs are normalized between 0 and 1?

t-rutten:
I added a prediction after training--thanks for the suggestions!

Issue #52, opened 2021-04-13
Convert examples to Live Markdown


seanmor5:
Notebooks are an excellent teaching resource. Examples should be converted to: [Livebook's](https://github.com/elixir-nx/livebook) Live Markdown with accompanying explanations of the models demonstrated.

Issue #25, opened 2021-03-30
Integration with common dataset format


seanmor5:
Requires developments in https://github.com/elixir-nx/nx/issues/301 and whatever external efforts emerge.

t-rutten:
Would it be worth shifting or adding function clauses to the current `Axon.Training` API to be more friendly to `inputs` and `targets` that are streams? This would seem to entail swapping `Enum` functions (like `zip/2` and `with_index/1`) for their `Stream` versions. 

Additionally, we could avoid calling `Enum.count/1` in `Training.train_epoch/5` by requiring inputs and labels to each be a tuple of an enumerable and metadata about the enumerable--like `{data, %{total_batches: _}} = _inputs`.

seanmor5:
@t-rutten I think we should go ahead and make the switch to use streams in the training API instead of Enumerables. I think any dataset format will logically have to be implemented as a stream anyway to avoid having to load large datasets entirely into memory.

As far as avoiding usage of `Enum.count`, I think we can instead just change the average calculation to use the current training step, that way we can support cases where the size of the dataset is unknown ahead of time and where users are specifying `steps` per epoch

seanmor5:
With the new training API, I think we're pretty much set on allowing Axon to work with anything that implements `Enumerable`, which I think is flexible enough as is.

PR #53, opened 2021-04-14
Datasets reorganization

t-rutten:
Here we've moved code for downloading and handling the CIFAR, MNIST, and FashionMNIST datasets used by the example scripts into separate modules under `Axon.Data`.

Most of the change are reorganization and de-duplication of code for downloading and processing datasets, but we also add options to change the batch size as well as the directory where datasets are stored. This allows for calls like

```elixir
Axon.Data.MNIST.download_images(data_path: 'tmp/my_data/mnist', batch_size: 200)
```

The API for `MNIST` and `FashionMNIST` looks like
```elixir
# Fetch images and labels together
{train_images, train_labels} = Axon.Data.MNIST.download()

# Or get them separately
train_images = Axon.Data.MNIST.download_images()
train_labels = Axon.Data.MNIST.download_labels()
```

Right now `Axon.Data.CIFAR` only supports `download/1` which returns the image/label tuple.

I'm not sure exactly where these modules should be located in the repo--I'd greatly appreciate feedback on that and any other suggestions!

### DIFF
--- a/examples/cifar10.exs
+++ b/examples/cifar10.exs
@@ -1,92 +1,3 @@
-defmodule CIFAR do
-  defp unzip_cache_or_download(zip) do
-    base_url = 'https://www.cs.toronto.edu/~kriz/'
-    path = Path.join("tmp", zip)
-
-    data =
-      if File.exists?(path) do
-        IO.puts("Using #{zip} from tmp/\n")
-        File.read!(path)
-      else
-        IO.puts("Fetching #{zip} from https://www.cs.toronto.edu/~kriz/cifar-10-binary.tar.gz\n")
-        :inets.start()
-        :ssl.start()
-
-        {:ok, {_status, _response, data}} = :httpc.request(:get, {base_url ++ zip, []}, [], [])
-        File.mkdir_p!("tmp")
-        File.write!(path, data)
-
-        data
-      end
-
-    data
-  end
-
-  def parse_images(content) do
-    {imgs, labels} =
-      for <<example::size(3073)-binary <- content>>, reduce: {[], []} do
-        {imgs, labels} ->
-          <<label::size(8)-bitstring, image::size(3072)-binary>> = example
-          # In order to "stack" we need a leading batch dim,
-          # we can introduce an Nx.stack for this similar to
-          # torch.stack
-          img =
-            image
-            |> Nx.from_binary({:u, 8})
-            |> Nx.reshape({1, 3, 32, 32})
-
-          label =
-            label
-            |> Nx.from_binary({:u, 8})
-
-          {[img | imgs], [label | labels]}
-      end
-
-    {Nx.concatenate(imgs, axis: 0), Nx.concatenate(labels, axis: 0)}
-  end
-
-  def download(zip) do
-    gz = unzip_cache_or_download(zip)
-
-    # Extract image and labels from tarfile, ideally we can also do this lazily
-    # such that we overlap training with IO - essentially have workers reading
-    # batches from the files and caching them in memory while the accelerator
-    # is busy doing actual training
-    with {:ok, files} <- :erl_tar.extract({:binary, gz}, [:memory, :compressed]) do
-      {imgs, labels} =
-        files
-        |> Enum.filter(fn {fname, _} -> String.match?(List.to_string(fname), ~r/data_batch/) end)
-        |> Enum.map(fn {_, content} -> Task.async(fn -> parse_images(content) end) end)
-        |> Enum.map(&Task.await(&1, :infinity))
-        |> Enum.unzip()
-
-      # Batch and one hot encode
-      #
-      # Ideally, we have something we can use as a batched stream
-      # such that we can lazily query for batches and avoid loading the
-      # entire dataset into memory, with this we could also apply lazy
-      # transformations enabling things like data echoing, see TFDS
-      imgs =
-        imgs
-        |> Nx.concatenate(axis: 0)
-        # We shouldn't have to do this, the batch should wrap or truncate
-        |> Nx.slice([0, 0, 0, 0], [49984, 3, 32, 32])
-        |> Nx.divide(255)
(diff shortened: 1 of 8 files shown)


josevalim:
What if move this outside of Axon into a separate datasets library? 🤔 For example, I can see using datasets in dev/test but there is no need to have it in prod.

seanmor5:
I agree with @josevalim I really like the idea of having this in a separate library. Something similar to Torchvision, Torchtext, etc. It looks like the framework is already mostly there as well for downloading the datasets.

t-rutten:
A separate library makes sense. Should it be inside the Axon repo, comparable to e.g. `exla` inside the Nx repo but alongside the `nx` project, or totally separate?

josevalim:
A totally separate project. :) we will also move EXLA and Torchx outside of Nx at some point once they become more stable.

btw, should we return binaries on download or should we return tensors? I guess binaries are fine and more general purpose?

nyo16:
> A totally separate project. :) we will also move EXLA and Torchx outside of Nx at some point once they become more stable.
> 
> btw, should we return binaries on download or should we return tensors? I guess binaries are fine and more general purpose?
The only maybe reason to have it in binaries is if you want to pass the images to another lib for image processing (if needed).

t-rutten:
Returning tensors would be nice for immediate use with Axon and Nx: the user wouldn't have to worry about the shape of the dataset to process the binaries.

Image processing with another library would certainly be a reason to keep them as binaries or in a different format. @nyo16 do you have any elixir libs in mind? [Torchvision.datasets](https://pytorch.org/vision/stable/datasets.html#mnist) returns [PIL](https://pillow.readthedocs.io/en/stable/) images and conversion to tensors happens with a transform function. 

We could do something similar here by returning a tensor--converted from binary and with a `transform/1` option applied to it. Changes would be necessary as progress is made on #25.

josevalim:
Could we make it return `{binary, type}`? So we just need to unpack it and call `Nx.from_binary`?

t-rutten:
Maybe `{binary, type, shape}`?

josevalim:
Sounds good to me!

nyo16:
> Returning tensors would be nice for immediate use with Axon and Nx: the user wouldn't have to worry about the shape of the dataset to process the binaries.
> 
> Image processing with another library would certainly be a reason to keep them as binaries or in a different format. @nyo16 do you have any elixir libs in mind? [Torchvision.datasets](https://pytorch.org/vision/stable/datasets.html#mnist) returns [PIL](https://pillow.readthedocs.io/en/stable/) images and conversion to tensors happens with a transform function.
> 
> We could do something similar here by returning a tensor--converted from binary and with a `transform/1` option applied to it. Changes would be necessary as progress is made on #25.

I think this would be great. My cases are around augmentation of the image(s). I don't have a specific lib in mind.
I like your idea to pass an optional transformation.

seanmor5:
@nyo16 With respect to augmenting images, you might be interested in: https://github.com/4rtemi5/imax

It is a pure Jax implementation of image augmentations and transformations. Theoretically, a library like that could be written entirely in `Nx` defn and then executed with something like EXLA to achieve the same effect without having to shell out to PIL or another library. Although, I assume PIL and others probably have more efficient implementations.

t-rutten:
I've moved the changes in this PR to a different repo: https://github.com/t-rutten/axon_datasets. 
`download/1` returns a tuple like `{binary_data, tensor_type, shape}` and accepts transformation options. Further suggestions are totally welcome.

Per #52, I'd be happy to port current examples to Live Markdown files that utilize AxonDatasets via `Mix.install/2` if that would be helpful.

josevalim:
Thanks @t-rutten! @seanmor5, should we have them inside the elixir-nx organization? Are you ok with this too, @t-rutten? Also, I am thinking we don't need `Axon` in the name, as they should be general purpose?

seanmor5:
@josevalim If it is okay with @t-rutten I think it makes sense to have them in the organization so it's more visible. I would also drop Axon from the name so they are more general purpose for other ML libraries

josevalim:
Btw, the package name "datasets" is taken though, any suggestions?

t-rutten:
Sure thing, let's move the repo to the organization.

NxDatasets? NxData? Nx is more general purpose than Axon...

josevalim:
Yeah but this isn't Nx related either. :D Which kinda of datasets are we going to keep? Maybe we can use something like that for the name?

batate:
ExDatasets?

Short, easy to type, shaped for Elixir consumption.

batate:
Good for SEO: DataScienceSets?

Short and descriptive, but not overly descriptive: Corpus? ExCorpus?

-bt

t-rutten:
We can start with computer vision related datasets and add text ones later (I do like ExCorpus @batate).

Maybe ExVision? VisionData? ImageData?

batate:
ExCorpus.* seems pretty good to me. 

ExCorpus.Vision
ExCorpus.Stats
ExCorpus.xxx

seems like a good start.

t-rutten:
I feel like "corpus" denotes language and text. @batate maybe your first suggestion, ExDatasets, is best: ExDatasets.Image, ExDatasets.Text, ExDatasets.Audio.

josevalim:
Not a big fan of `Ex` prefixes... :D Based on your suggestions, I could think of: science_data / scidata / scisets / scidatasets / sci_datasets. I like scidata/Scidata.

batate:
scidata is excellent.

t-rutten:
I like scidata!

josevalim:
Perfect! @t-rutten, I have created a repo in this organization, alongside a team, and gave you access: https://github.com/elixir-nx/scidata

Feel free to push or submit a PR to the scidata repo accordingly. :)

t-rutten:
Great, thanks @josevalim!

josevalim:
@t-rutten I pushed an initial commit with the LICENSE and made sure you have the proper write permissions. :)

PR #54, opened 2021-04-17
Initial custom layer implementation

### LINKED ISSUE #35: Support custom layers
Modularity of layers is easy because we can use regular Elixir functions, but we need a solution for specifying and using custom trainable parameters in layers.

seanmor5:
Resolves #35 WIP, but this is how custom layers will look, specifically look at the xor example for the worst custom layer implementation of all time using `atan2` and a trainable weight. I will clean this up and fix the tests before this can be merged.

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -83,9 +83,15 @@ defmodule Axon do
   @doc """
   A custom Axon layer.
   """
-  def layer(parent, op, output_shape, parameters, opts \\ []) do
-    {name, opts} = Keyword.pop(opts, :name)
+  def layer(parent, op, output_shape, parameters, name \\ nil, opts \\ [])
+      when is_atom(op) or is_function(op) do
+    op = if is_atom(op), do: op, else: :layer
+
     {id, name} = unique_identifiers(op, name)
+
+    parameters =
+      parameters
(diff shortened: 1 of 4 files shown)


lib/axon.ex
```diff
@@ -80,6 +80,50 @@ defmodule Axon do
   @doc false
   defstruct [:id, :name, :output_shape, :parent, :op, :params, :opts]
 
+  @doc """
+  A custom Axon layer.
+  """
+  def layer(parent, op, output_shape, parameters, name \\ nil, opts \\ [])
+      when is_atom(op) or is_function(op) do
+    op = if is_atom(op), do: op, else: :layer
```
josevalim, line 88: We need a bit more docs. :)

In any case, when is `op` an atom? Is that only for Axon own layers?

josevalim:
It looks absolutely great, I have dropped just one tiny comment. :)

lib/axon.ex
seanmor5: Yes, but I think with this change I can greatly simplify the compiler which may mean that it is always a function or perhaps not, there are just a few things I need to think about first.

josevalim: Yeah, I think normalizing is good and then have the “op” as an option to optionally name it?

PR #55, opened 2021-04-18
Format code

AlexJuca:
This PR formats a section of the flatten function so as  to maintain consistency across the codebase.


seanmor5:
Hi @AlexJuca thanks for the PR! I am about to merge a PR #54 that will conflict with this branch because it makes a ton of changes.

PR #56, opened 2021-04-19
code improvements

AlexJuca:
This PR contains the following changes and improvements:

- Cleans up documentation, artifacts of a previous conflict were lingering in the docs.
- Remove duplicate calls to System.unique_integer and extracted to single more contextual unique_id function.

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -81,11 +81,6 @@ defmodule Axon do
   defstruct [:id, :name, :output_shape, :parent, :op, :params, :opts]
 
   @doc """
-<<<<<<< Updated upstream
-=======
-<<<<<<< Updated upstream
-=======
->>>>>>> Stashed changes
   A custom Axon layer.
   """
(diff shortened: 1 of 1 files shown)


lib/axon.ex
```diff
@@ -1136,10 +1127,12 @@ defmodule Axon do
     end
   end
 
+  defp unique_id(modifiers) when is_list(modifiers), do: System.unique_integer(modifiers)
```
seanmor5, line 1130: I don't think this is really all that necessary because it's an extra call to do the same thing without avoiding any duplication

seanmor5 (commented):
I will fix the docs upstream, thanks!

AlexJuca:
Go ahead @seanmor5

lib/axon.ex
AlexJuca: Thanks, I'll try and add something with more value next time.

PR #60, opened 2021-04-28
Training API uses streams

t-rutten:
Per discussion in #25, changes here adapt the training API to accept `inputs` and `labels` that are streams.

### DIFF
--- a/lib/axon/training.ex
+++ b/lib/axon/training.ex
@@ -144,45 +144,48 @@ defmodule Axon.Training do
   ## Helpers
 
   defp train_epoch(step_fn, model_state, inputs, targets, epoch, opts) do
-    total_batches = Enum.count(inputs)
-
     {log_every, jit_opts} = Keyword.pop(opts, :log_every)
 
     log_freq =
-      case log_every do
-        :epoch ->
-          total_batches
-
-        x when is_integer(x) ->
-          x
-
-        :none ->
-          total_batches + 1
+      if is_integer(log_every) do
+        log_every
+      else
+        -1
       end
 
     dataset =
       inputs
-      |> Enum.zip(targets)
-      |> Enum.with_index()
+      |> Stream.zip(targets)
+      |> Stream.with_index()
 
-    for {{inp, tar}, i} <- dataset, reduce: {model_state, Nx.tensor(0.0)} do
-      {model_state, state} ->
-        {model_state, batch_loss} = Nx.Defn.jit(step_fn, [model_state, inp, tar], jit_opts)
+    {model_state, avg_loss, total_batches} =
+      for {{inp, tar}, i} <- dataset, reduce: {model_state, Nx.tensor(0.0), 0} do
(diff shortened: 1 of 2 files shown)


lib/axon/training.ex
```diff
@@ -144,45 +147,49 @@ defmodule Axon.Training do
...
-          total_batches
+    dataset =
+      inputs
+      |> Stream.zip(targets)
+      |> Stream.with_index()
 
-        x when is_integer(x) ->
-          x
+    {model_state, avg_loss, total_batches} =
+      for {{inp, tar}, i} <- dataset, reduce: {model_state, Nx.tensor(0.0), 0} do
```
seanmor5, line 158: I'm guessing this runs to the end of the stream? I think perhaps we'll want an option to terminate before the end of a stream. I'm trying to think of the best way to do that without loading `n` steps all into memory at once with something like `Enum.take`, any thoughts?

seanmor5 (commented):
Thanks for putting this together! I left a comment to see if we can address that before we merge. BTW, make sure to run the formatter so it passes the CI!

seanmor5:
> Thanks for putting this together! I left a comment to see if we can address that before we merge. BTW, make sure to run the formatter so it passes the CI!

Ahh I just realized the formatting failure is probably from my recent commit, in any case, feel free to run the formatter on everything so it passes before we merge :)

lib/axon/training.ex
josevalim: You can do a `Stream.take(n)` or do a throw/try+catch.

t-rutten: Can we go with `Enum.reduce_while/3`? We just need the accumulator when processing each batch, but does that function keep the accumulator as well as the first _n_ elements of the stream in memory at once? With the continue/halting function of `reduce_while` it would easy to use incorporate different early stopping criteria.

t-rutten: @seanmor5 @josevalim, do you want me to add options for early stopping in this PR? Otherwise it'll be a straightforward addition to convert the comprehension to `reduce_while` later on.

[Early stopping](https://pytorch-lightning.readthedocs.io/en/latest/common/early_stopping.html#stopping-an-epoch-early) might be specified by a function that accepts a combination of current batch/step loss, previous batch loss, average loss, and batch index and returns a boolean indicating whether training should continue.

seanmor5: @t-rutten I would hold off on that for now, I would like to implement that as a callback when those are added into the training API

t-rutten: Makes sense @seanmor5. Do you have any more suggestions for the changes here?

seanmor5: No changes, I'm happy with where this is, unless you have anything else you'd like to add, I'll merge :)

seanmor5: No changes, I'm happy with where this is, unless you have anything else you'd like to add, I'll merge :)

PR #61, opened 2021-04-30
Add recurrent layers

### LINKED ISSUE #24: Add Recurrent Layers
Require solution for managing layers that maintain state / return multiple outputs

- [ ] gru
- [ ] lstm
- [ ] conv_lstm

seanmor5:
Resolves #24 

Everything works for now, we just need to integrate RNNs into the high-level API. Unfortunately static_unrolls are going to be really expensive in terms of grad and compilation, so for long sequences it's not feasible to use, but it will work for short sequences as a proof of concept until we can add something  like `Nx.scan`


josevalim (commented):
Nice!

PR #66, opened 2021-05-22
feat: add orthogonal initializer

polvalente:
This PR aims to add `Axon.Initializers.orthogonal/1` for #1

From what I read in https://www.tensorflow.org/api_docs/python/tf/keras/initializers/Orthogonal, we can implement this through Nx.LinAlg.qr + Nx.reshape in some combination.

@seanmor5 is this the right path? I've only added this base case as a proof of concept

### DIFF
--- a/lib/axon/initializers.ex
+++ b/lib/axon/initializers.ex
@@ -577,16 +577,14 @@ defmodule Axon.Initializers do
   Initializes a tensor with an orthogonal distribution.
 
   For 2-D tensors, the initialization is generated through the QR decomposition of a random distribution
-  For tensors with more than 2 dimensions, a 2-D tensor with shape `{shape[0] * shape[1] * ... shape[n-2], shape[n-1]}`
+  For tensors with more than 2 dimensions, a 2-D tensor with shape `{shape[0] * shape[1] * ... * shape[n-2], shape[n-1]}`
   is initialized and then reshaped accordingly.
 
   ## Options
 
     * `:shape` - output shape. Must be at least rank `2`
-    * `:type` - output type. Defaults to `{:f, 32}`
-    * `:scale` - scale of the output distribution. Defaults to `1.0e-2`
-    * `:distribution` - output distribution. One of `:normal`, `:truncated_normal`,
-      or `:uniform`. Defaults to `:normal`
+    * `:type` - random seed's type. Defaults to `{:f, 32}`
+    * `:distribution` - output distribution. One of [`:normal`, `:uniform`]. Defaults to `:normal`
 
   ## Examples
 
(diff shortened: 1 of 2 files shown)


lib/axon/initializers.ex
```diff
@@ -573,6 +573,49 @@ defmodule Axon.Initializers do
...
+        u8
+        1
+      >
+  """
+  defn orthogonal(opts \\ []) do
+    opts = keyword!(opts, [:shape, type: {:f, 32}, distribution: :normal])
+
+    assert_greater_equal_rank!(opts[:shape], 2)
+
+    random_seed = Nx.random_normal(opts[:shape])
```
polvalente, line 610: The idea is that we can generate a random seed with a distribution dependent on `opts[:distribution]` with the shape as described in the docstring. 

I haven't handled this yet for simplicity in the draft, but I plan on adding these if this is the correct path for the PR :)

lib/axon/initializers.ex
```diff
@@ -573,6 +573,49 @@ defmodule Axon.Initializers do
...
+      iex> Nx.type(t)
+      {:f, 32}
+      iex> Nx.shape(t)
+      {3, 3}
+      iex> identity = t |> Nx.transpose() |> Nx.dot(t)
+      iex> Nx.all_close?(identity, Nx.eye(t), atol: 1.0e-3, rtol: 1.0e-3)
+      #Nx.Tensor<
+        u8
+        1
+      >
```
polvalente, line 603: I wasn't sure on how to test this since the matrices are random. However, since it's guaranteed to be orthogonal, the property that the transpose is it's inverse should hold

josevalim: It probably makes sense to move those to actual test files so we can start building the testing infrastructure.

josevalim: But testing the property is a neat idea!

lib/axon/initializers.ex
```diff
@@ -573,6 +573,85 @@ defmodule Axon.Initializers do
...
+            m = Enum.reduce(tl(tuple_list), 1, &(&1 * &2))
+            {m, n}
+          else
+            shape
+          end
+
+        random_seed =
+          case distribution do
+            :uniform ->
+              orthogonal_random_uniform(flat_shape, type)
```
seanmor5, line 627: I don't think there's a need for this extra level of indirection

lib/axon/initializers.ex
```diff
@@ -573,6 +573,85 @@ defmodule Axon.Initializers do
...
+
+            :normal ->
+              orthogonal_random_normal(flat_shape, type)
```
seanmor5, line 630: Same as above

seanmor5 (approved):
This is great @polvalente thank you! I just left 2 minor comments

lib/axon/initializers.ex
polvalente: If I don't do it this way, I get: `** (ArgumentError) defn must return a tensor expression or a tuple, got: #Nx.Tensor`

Another way to do this is:

```elixir
case distribution do
            :uniform ->
              Nx.random_uniform(flat_shape, type: type, backend: Nx.Defn.Expr)

            :normal ->
              Nx.random_normal(flat_shape, type: type, backend: Nx.Defn.Expr)

            dist ->
              raise ArgumentError,
                    "invalid distribution #{inspect(dist)} passed to orthogonal/1"
          end
```

Which one do you prefer?

josevalim: The issue is that you are calling this code inside the transform and the transform does not follow defn rules. Thats why you need to call a defnp that calls `Nx.random_...`. Or indeed you need to pass the `backend` option.

seanmor5: I think I like the backend way better :)

polvalente: Fine by me :)
I'll push it in a few minutes

PR #68, opened 2021-05-24
Use parameter map

### LINKED ISSUE #58: Parameters declared out of order initialize incorrectly
This stems from our current (fragile) way of ensuring parameters are in the correct order by generating a unique ID for each parameter like:

```elixir
System.unique_integer([:positive, :monotonic])
```

and then sorting on the ID because they are guaranteed to be ordered. Internally this isn't that much of a problem, but with the addition of custom layers, it's possible to run into the following case:

```elixir
bias = Axon.param("bias", {}, initializer: :zeros)
weight = Axon.param("weight", {}, initializer: :ones)

Axon.layer(x, fn x, w, b -> Nx.add(Nx.multiply(x, w), b) end, output_shape, [weight, bias])
```

During initialization, `weight` will be initialized as `bias` and `bias` will be initialized as `weight`. When they are the same shapes as in this instance, I think this will lead to silent bugs. We could document, but I believe our current method is fragile and needs to be refactored altogether.

We need a better way to track parameters at each layer. This is really an issue in the sense that we can't just declare a parameter and use it in an operation like:

```elixir
def dense(%Axon{output_shape: shape} = x, units) do
  w = Axon.param("weight", {elem(shape, 1), units)
  b  = Axon.param("bias", {1, units})
  Nx.add(Nx.dot(x, w), b)
end
```

So we need a better way to ensure parameters are initialized and used in the correct places. I believe the best option is to ensure each parameter has a unique name, and then pass `params` as a map to `defn` - which would make this an upstream issue.

seanmor5:
Resolves #58 and aims to simplify parameter handling logic in custom layers and in other areas as well.

Custom layers are now declared like:

```elixir

weight = Axon.param("weigh", {})

Axon.layer(x, {}, fn x, params -> Nx.arctan2(params["weight"], x) end, %{"weight" => weight})
```

The keys passed to layer are the same as those used to access parameters in `params`. This breaks pretty much every test though, and a lot of refactoring needs to be done


josevalim:
Quick question: binary or atom keys?

seanmor5:
@josevalim I went with binary keys because of parameter naming conventions in other frameworks. For example, in Keras parameters are usually namespaced like:

`dense_1/kernel:0`

While we can represent that as an atom I figured it would be easier and more natural with binaries. The keys used to access parameters during the model's forward pass are namespaced and guaranteed unique. So in built-in layers we can access their parameters directly from the name field of the parameter struct. For custom layers, it's impossible to know the unique name at the time of parameter creation, which is what the parameter map is for. Internally the compiler resolves `param["weight"]` to the correct parameter.

PR #71, opened 2021-06-03
Add training callbacks

seanmor5:
WIP


lib/axon/metrics.ex
```diff
@@ -220,16 +220,16 @@ defmodule Axon.Metrics do
...
   ## Argument Shapes
-
+  
     * `y_true` - $\(d_0, d_1, ..., d_n\)$
     * `y_pred` - $\(d_0, d_1, ..., d_n\)$
-
+  
   ## Examples
-
+
```
josevalim: You probably don't want to add all of these newlines? :D

lib/axon/training.ex
```diff
@@ -229,89 +289,130 @@ defmodule Axon.Training do
       init_fn = &init_values/0
 
   """
-  def train({init_fn, step_fn}, inputs, targets, opts \\ []) do
+  def train(
+        %Step{init: init_fn, step: step_fn, callbacks: callbacks},
```
josevalim, line 293: 😍

PR #72, opened 2021-06-05
Add resize layer

seanmor5:
Proof-of-concept of basic upsampling layer using bilinear interpolation. Requires https://github.com/elixir-nx/nx/issues/419 to be fixed. Needs to be tested on EXLA. The current `slice_grid` implementation might result in issues because it generates a large expression with a bunch of slices.

### DIFF
--- a/lib/axon/layers.ex
+++ b/lib/axon/layers.ex
@@ -1748,9 +1748,15 @@ defmodule Axon.Layers do
           ]
         ]
       >
+
+  ### Error cases
+
+      iex> img = Nx.iota({1, 1, 3, 3}, type: {:f, 32})
+      iex> Axon.Layers.resize(img, shape: {4, 4}, method: :foo)
+      ** (ArgumentError) invalid resize method :foo, resize method must be one of :nearest
   """
   defn resize(img, opts \\ []) do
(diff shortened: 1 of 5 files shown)


lib/axon/layers.ex
```diff
@@ -1306,4 +1306,80 @@ defmodule Axon.Layers do
     new_shape = transform(Nx.shape(x), &Axon.Shape.flatten/1)
     Nx.reshape(x, new_shape)
   end
+
+  @doc """
+  Upsamples image.
```
josevalim, line 1311: This is generally used by GANs, right?

```suggestion
  Upsamples image.
  
  This can be used by GAN (Generative Adversarial Networks)
  to construct an image our of a random vector sample mimicking
  an image from a real distribution.
```

josevalim:
> The current slice_grid implementation might result in issues because it generates a large expression with a bunch of slices.

Is there a more efficient way to implement this on EXLA? 🤔

seanmor5:
@josevalim I think with a gather, but I'm also watching this issue: https://github.com/google/jax/issues/6872

I think it would be worth a benchmark to see which is more efficient in this case

Issue #59, opened 2021-04-19
Create mechanism for easy model composition


seanmor5:
For now, we'll only consider how this should work in the model creation and execution API, but it will touch the training API as well.

Consider the models in a basic GAN:

```elixir
generator =
  Axon.input({nil, 100})
  |> Axon.dense(128, activation: :tanh)
  |> Axon.dense(512, activation: :tanh)
  |> Axon.dense(784, activation: :tanh)
  |> Axon.reshape({1, 28, 28})

discriminator =
  Axon.input({nil, 1, 28, 28})
  |> Axon.dense(128, activation: :relu)
  |> Axon.dense(1, activation: :sigmoid)
```

In order to train, what you'd want to do is something like:

```elixir
combined = compose(discriminator, generator)  # represents D(G(input)) 
step_d = Axon.Training.step(discriminator, :binary_cross_entropy, Axon.Optimizers.sgd(0.005)
step_g = Axon.Training.step(combined, :binary_cross_entropy, Axon.Optimizers.adam(0.01)
```

And then you can alternate using `step_d` and `step_g` to train on valid / fake images. Unfortunately, we currently don't support model composition in this sense - you can define functions `generator` and `discriminator` without an input block, but there's no way to cleanly determine which parameters belong to which model. Ideally, you'd be able to compose models in some way so that when you initialize, predict, train, etc. parameters are grouped:

```elixir
combined = compose(discriminator, generator)
{d_params, g_params} = combined_params = Axon.init(combined)
Axon.predict(combined, combined_params)

{{d_params, g_params}, _} =
  combined
  |> Axon.Training.step(:binary_cross_entropy, Axon.Optimizers.adam(0.01)
  |> Axon.Training.train(inputs, targets)
```

Whatever the implementation is, it will involve adding some metadata to parameters to express that expresses their ownership to a given model. From an API perspective, one option is to introduce `Axon.compose` for composing Axon structs into a single model while preserving parameter information, although I'm not sure I love that right now.

seanmor5:
I've been experimenting a bit and after starting #81 I believe I have a solution to this issue. Introduce `Axon.function`. The idea is that `Axon.function` takes a block of layers with inputs and returns an anonymous function with arity matching the number of inputs in the block. So the GAN would look like:

```elixir
generator =
  Axon.input({nil, 100})
  |> Axon.dense(128, activation: :tanh)
  |> Axon.dense(512, activation: :tanh)
[...]

PR #81, opened 2021-06-20
Initial draft of hooks

### LINKED ISSUE #77: Add training hooks
See: https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.register_forward_hook
And: https://pytorch.org/docs/stable/_modules/torch/nn/modules/module.html#Module.register_backward_hook

seanmor5:
Resolves #77 when merged.

This is an initial draft but it motivates 2 discussions: host callbacks and reusable model blocks.

In order to perform side-effecting operations with hooks, Nx needs to support host callbacks. Once that's done, it likely means that hooks will always be implemented as host callbacks, but the implementation will need to evolve once we support them. I don't see the API being impacted much once we introduce host callbacks.

As for model blocks: In PyTorch has two kinds of hooks - tensor hooks and module hooks. In Axon we're more focused on module-style hooks for now. Module hooks have the following signature: `hook(module, input, output) :: None`. The equivalent in Axon would be: `hook(axon_struct, input, output) :: nil`. For both forward and backward hooks, even if they are only meant to be side-effecting operations, I'd prefer they always return an output such that after a layer the output is basically what's shown in this PR.

One issue is that we're currently unable to support hooks registered on large blocks of a model. For example, if I have a block of Convs:

```elixir
block =
  x
  |> conv()
  |> conv()
  ...
```

and I tried to register a hook, I would only have access to the input to the last conv in the block, not `x`. This motivates introducing a module-like abstraction which allows you to treat groups of layers as a single block. This is essentially the same issue we have in #59 as well.

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -78,7 +78,7 @@ defmodule Axon do
   @type t :: %__MODULE__{}
 
   @doc false
-  defstruct [:id, :name, :output_shape, :parent, :op, :params, :policy, :opts]
+  defstruct [:id, :name, :output_shape, :parent, :op, :params, :policy, :on_forward, :on_backward, :opts]
 
   @doc """
   Custom Axon layer with given parent.
@@ -126,7 +126,9 @@ defmodule Axon do
       op: op,
       params: parameters,
       policy: Axon.MixedPrecision.create_policy(),
-      opts: opts
+      opts: opts,
+      on_forward: nil,
+      on_backward: nil
     }
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:
Looks great! Just one note: I think even with host callbacks, you would want to choose if you want to bring the data back or the hook is a Nx transformation.

seanmor5:
I will have a new PR for this shortly integrated with Nx's new hook API

PR #86, opened 2021-07-08
Add global pooling layers

seanmor5:


test/axon_test.exs
```diff
@@ -477,6 +477,20 @@ defmodule AxonTest do
...
+    test "works with options" do
+      for pool <- @global_pooling_layers do
+        assert %Axon{name: "pool", output_shape: {nil, 1}} =
+                 apply(Axon, pool, [
+                   Axon.input({nil, 1, 28, 28}),
+                   [keep_axes: false, name: "pool"]
+                 ])
+      end
+    end
+  end
```
josevalim, line 492: This doesn't test it actually works, right? It may be worth changing those tests later on to execute, even if we don't assert on the result, but to guarantee Nx+EXLA compiles it. :)

josevalim (approved):
LGTM!

test/axon_test.exs
seanmor5: Correct, I have decided that to handle this here and with other layers I am going to introduce a `compiler_test.exs` file. So `axon_test` will assert the model is being constructed and inspected correctly and compiler test will assert on results

PR #88, opened 2021-07-14
Add compiler test

seanmor5:
The purpose of this is to add some basic tests of compilation of each layer before building up more complex integration tests. The focus is mainly to make sure that initialization and forward pass, as well as more complex things like model freezing and parameter policies work as expected. I'm pretty much testing models exactly against their implementation in `Axon.Layer` as the implementations should be tested individually in the layer test.

### DIFF
--- a/test/compiler_test.exs
+++ b/test/compiler_test.exs
@@ -440,7 +440,8 @@ defmodule CompilerTest do
     end
 
     test "initializes with custom initializers" do
-      model1 = Axon.input({nil, 3, 32, 32}) |> Axon.conv(32, name: "conv", kernel_initializer: :zeros)
+      model1 =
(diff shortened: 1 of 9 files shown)


test/compiler_test.exs
```diff
@@ -0,0 +1,429 @@
...
+      model =
+        Axon.input({nil, 32})
+        |> Axon.dense(1, name: "dense")
+        |> Axon.freeze()
+
+      assert {init_fn, predict_fn} = Axon.compile(model)
+
+      backward = fn params, input ->
+        Nx.Defn.grad(params, &Nx.mean(predict_fn.(&1, input)))
+      end
```
seanmor5, line 125: This could possibly be included as a function of the model similar to `Axon.predict` and `Axon.init`. Something like:

```elixir
model = ...
loss_fn = ...
gradients = Axon.backward(model, params, input, loss_fn)
```

josevalim (approved):
Beautiful!

PR #106, opened 2021-08-30
Fix compilation cache

### LINKED ISSUE #89: Cache results instead of functions
Currently we cache branches of the graph such that functions aren't built more than once. This means functions will be re-used, but they will be called mutliple times rather than their result being re-used. For example, RNNs will be unrolled 3(!!!!) times on the forward pass in the current implementation. This is a significant performance issue.

seanmor5:
Resolves #89

### DIFF
--- a/lib/axon/compiler.ex
+++ b/lib/axon/compiler.ex
@@ -113,14 +113,14 @@ defmodule Axon.Compiler do
 
     fn params, inputs ->
       inputs = maybe_flatten(inputs)
-      {fun, _} = to_predict_fun(graph, %{}, input_map)
+      {expr, _} = to_predict_fun(graph, %{}, input_map, params, inputs)
 
-      case fun do
-        [_ | _] = funs ->
-          do_recur_apply(funs, params, inputs, [])
+      case expr do
+        [_ | _] = exprs ->
+          do_recur_to_tuple(expr, [])
 
-        fun when is_function(fun) ->
-          fun.(params, inputs)
+        expr ->
+          expr
       end
     end
   end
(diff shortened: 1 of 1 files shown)


lib/axon/compiler.ex
```diff
@@ -218,21 +218,22 @@ defmodule Axon.Compiler do
...
-        {fun, cache}
+      %{^id => res} ->
+        {res, cache}
 
       %{} ->
-        {fun, cache} = recur_predict_fun(graph, cache, input_map)
-        cache = Map.put(cache, id, fun)
-        {fun, cache}
+        {fun, cache} = recur_predict_fun(graph, cache, input_map, params, inputs)
+        res = fun.()
```
josevalim, line 234: We can fully skip returning the `fun` and have it return `res` directly. :)

lib/axon/compiler.ex
```diff
@@ -113,14 +113,14 @@ defmodule Axon.Compiler do
 
     fn params, inputs ->
       inputs = maybe_flatten(inputs)
-      {fun, _} = to_predict_fun(graph, %{}, input_map)
+      {expr, _} = to_predict_fun(graph, %{}, input_map, params, inputs)
```
josevalim, line 116: As we discussed, the anonymous function will be stored in ETS, so we want to make all variables used inside the anonymous function are lightweight. So are we sure the `graph` and the `input_map` do not have large tensors in them?

seanmor5: `input_map` just maps layer names to indices in the `input` from `predict_fn(params, input)` so it will never contain large tensors. It is possible that the graph contains large tensors if the user embeds them in the graph with `Axon.constant`

josevalim (approved):

lib/axon/compiler.ex
seanmor5: Both `params` and `inputs` come from the top level function definition and will get a map and tuple of large tensors respectively - is that going to be problematic?

josevalim: Those are fine. We cache only the shapes!

PR #114, opened 2021-09-23
Fix: sample image shape

floatn:
Issue:
(ArgumentError) invalid input shape given to model, expected input with shape {nil, 1, 28, 28}, but got input with shape {1, 28, 28}


seanmor5:
Thanks!

PR #121, opened 2021-09-24
Fix: test images tesor shape

floatn:
Issue:
(ArgumentError) cannot reshape, current shape {3, 3, 32, 32} is not compatible with new shape {3, 28, 28}


seanmor5:
Thanks!

PR #125, opened 2021-09-30
Update examples to use Mix install - Ref1

floatn:
Refactoring of #122


seanmor5:
Great work and thank you! I will do a more in-depth review tomorrow, but at a quick glance this looks great :)

PR #126, opened 2021-10-05
Fixed reported epoch number in metrics logging (off-by-one)

arpieb:
There was an off-by-one error in the metrics logging output; fixed!


seanmor5 (approved):
Thanks!

Issue #101, opened 2021-08-27
Add training loop fusion


seanmor5:
With the support of `while` loops in `Nx` and progress towards streaming data to the device, we should consider an API which allows for fusing the training loop into a single XLA computation. We will have to consider how this ties in with callbacks, hooks, etc.

Issue #80, opened 2021-06-19
Support multiple objectives in training API


seanmor5:
Given multiple outputs, or just in general, the ability to specify multiple loss functions is necessary to have in the training API.

seanmor5:
Closed in #132

Issue #79, opened 2021-06-19
Integrate learning rate schedules


seanmor5:
As is, there is no way to anneal learning rate during training.

Issue #78, opened 2021-06-19
Add inference function to callback API


seanmor5:
Related to: #75 

Callbacks should have access to the inference function. Example usage: Saving generated images during GAN training

PR #132, opened 2021-10-12
New training API

### LINKED ISSUE #22: Integrate validation and testing into training API
More than likely this can be almost identical to the PyTorch approach where we define `validation_step` and `test_step` and then we can optionally include validation in with training and then testing with a separate `Axon.test`

seanmor5:
This PR implements a new training API modeled after [PyTorch Ignite](https://pytorch.org/ignite/concepts.html) which is a little more thought out and extensible than the previous API. When finished, this PR will resolve #22, #78, and #80; and provide at least what I think will be an easier path forward for #25, #79, and #101.

The basic idea is that `Axon.Training` implements conveniences and functions for running loops or reductions in the same way that PyTorch Ignite does. A "loop" is a struct which controls the inner workings of the loop, currently it contains the fields: `process`, `metrics`, `handlers`. `process` is a "process" struct which represents the loop processing function. A loop processing function accepts the loop state and some data, and returns an updated process state. The process struct contains an init and step function for initializing and updating the process state respectively.

`metrics` is a map of metrics to attach to the process state. This means you can apply metrics on predictions, parameters, optimizer state, etc.

`handlers` is a map of events and their specific handlers. `handlers` are meant to be side-effecting, so they don't alter training state in anyway. This might change in the future.

The current API exposes a `run` function that runs a given loop. It doesn't make any assumptions about whether we're doing training, validation, testing, etc. Ignite provides examples of some of the other cool things you can do with their Engines (like processing datasets and more).

There's also a `step` function which creates a loop from a model, loss, and optimizer. I plan on adding additional loop factories in the future. Ideally, we change the name of this to something like `supervised_train_step` or just `train_step` and then we can have `validation_step`, etc. 

Training doesn't change all that much:

```elixir
model
|> Axon.Training.step(:categorical_cross_entropy, Axon.Optimizers.adam(5.0e-3)
|> Axon.Training.metric(:accuracy, "Accuracy")
|> Axon.Training.run()
```

But, now we can do things like:

```elixir
def test_model(model, model_state, {test_images, test_labels}) do
    process_fn = fn _, {inp, _} ->
      preds = Axon.predict(model, model_state, inp)
      %{predictions: preds, loss: Nx.tensor(0.0, backend: Nx.Defn.Expr)}
    end

    process = %Axon.Training.Process{init: init_fn, step: process_fn}
    loop = %Axon.Training.Loop{process: process} |> Axon.Training.metric(:accuracy, "Accuracy")
    Axon.Training.run(loop, Stream.zip(test_images, test_labels), compiler: EXLA)
end
```

The above example can be cleaned up even further with some additional factories and refining of the API. 

Now the `Axon.Training` API is less of a training API and more of a loop API to be used in training. Not sure if it makes sense to continue to call it `Axon.Training` anymore becasue it theoretically could be used outside of training, but not sure if it really matters.

This is still WIP as there are a lot of things I want to play with/add. I also plan on adding a number of integration training tests with this new API.

### DIFF
--- a/examples/cifar10.exs
+++ b/examples/cifar10.exs
@@ -6,19 +6,7 @@ Mix.install([
 ])
 
 # Configure default platform with accelerator precedence as tpu > cuda > rocm > host
-case EXLA.Client.get_supported_platforms() do
-  %{'TPU' => _} ->
-    Application.put_env(:exla, :clients, default: [platform: :tpu])
-
-  %{'CUDA' => _} ->
-    Application.put_env(:exla, :clients, default: [platform: :cuda])
-
-  %{'ROCM' => _} ->
-    Application.put_env(:exla, :clients, default: [platform: :rocm])
-
-  %{'Host' => _} ->
-    Application.put_env(:exla, :clients, default: [platform: :host])
-end
+EXLA.Client.set_preferred_platform(:default, [:tpu, :cuda, :rocm, :host])
 
 defmodule Cifar do
   require Axon
--- a/examples/fashionmnist_autoencoder.exs
+++ b/examples/fashionmnist_autoencoder.exs
@@ -6,19 +6,7 @@ Mix.install([
 ])
 
 # Configure default platform with accelerator precedence as tpu > cuda > rocm > host
-case EXLA.Client.get_supported_platforms() do
-  %{'TPU' => _} ->
-    Application.put_env(:exla, :clients, default: [platform: :tpu])
-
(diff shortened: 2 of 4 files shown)


josevalim:
@arpieb @seanmor5 regarding the name, what about `Axon.Loop`? Then you can use `Axon.Loop.training` to generate a training loop. Or you can call it `Axon.Step` with `Axon.Step.training`.

Issue #91, opened 2021-07-18
ConvLSTM shape logic is broken


seanmor5:
Currently the parameter shape calculations and output shape calculations are incorrect for the ConvLSTM, making it unusable

PR #136, opened 2021-10-15
Fix: ConvLSTM shape logic

floatn:
#91


josevalim:
@floatn can you show the test? It seems theere is something wrong with the do-block test?

floatn:
Hm. When I pull it from upstream, it gives me the error, but a fresh clone succeeds indeed. Sorry, looks like my mistake. I'll try to figure it out by myself.

PR #167, opened 2021-11-23
Allow user-defined values in the `step_state` to propagate across iterations

zeionara:
This change allows to take values which remain unchanged in comparison to the result of the previous iteration and pass them to the next iteration. It can be useful in cases, when custom loop handlers keep some data in training state (for instance, minimum or maximum loss).

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -823,8 +823,6 @@ defmodule Axon.Loop do
                           {:halt, {:halted, final_metrics_map, state}}
 
                         {:continue, state} ->
-                          max_iter = state.iteration
-
                           zero_metrics =
                             Map.new(metric_fns, fn {k, _} ->
                               {k, 0}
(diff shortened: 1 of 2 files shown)


seanmor5:
Thanks! I think this is valuable. Can you please write a test or 2 ensuring user defined values are actually preserved?

zeionara:
Added a couple of tests for checking that user-defined values in `step_state` are actually propagated, but somewhy there are still failed checks. Unfortunately, the proposed solution only deals with numerical data for user-defined values; if we try to pass a value of another type (string, for instance), it will fail. Actually I have an idea on how to allow passing values of any types, but it would require more modifications not only in the `axon` library, but in the `nx` dependency as well. So, it's unclear would it be appropriate or not.

seanmor5:
The failed checks seemed to be transient, thanks :D

PR #168, opened 2021-11-27
Bump examples with current Nx

seanmor5:
Updates all examples and uses new Nx container protocol. I'm not fully confident in the Loop and Loop State stuff working as expected with the new container protocol, so I need to write some tests under `jit` to make sure everything is going well before this is merged :)

All of the examples are working though.


polvalente (approved):

lib/axon/loop/state.ex
```diff
@@ -39,6 +39,9 @@ defmodule Axon.Loop.State do
   # a way in Nx/EXLA to mark `:times` as a static property which is
   # not to be touched at JIT time.
   @enforce_keys [:step_state]
+  @derive {Nx.Container,
+           containers: [:step_state, :epoch, :max_epoch, :iteration, :max_iteration, :metrics],
```
josevalim, line 43: So as a reminder, those are the values you want to make available inside the GPU.

lib/axon/loop/state.ex
```diff
@@ -39,6 +39,9 @@ defmodule Axon.Loop.State do
...
+           keep: [:times]}
```
josevalim, line 44: I don't think you want to keep times. If you keep times, it means every loop interaction has a different `time`, which means it will use a different EXLA cache key and compile it from scratch. You either want to mark it as a container and send it upstream or have the `times` variable be reset and then you roll it back to the current interation once you get the stream back.

josevalim: `EXLA.jit_cached?/3` and `EXLA.stream_cached?/3` are on Nx latest.

lib/axon.ex
```diff
@@ -82,6 +82,11 @@ defmodule Axon do
   @type t :: %__MODULE__{}
 
   @doc false
+  @derive {
+    Nx.Container,
+    containers: [],
+    keep: [:id, :name, :output_shape, :parent, :op, :params, :policy, :opts]
```
josevalim, line 88: Do you need to access all of those fields inside `defn`? If so, it is fine to keep. If not, they shouldn't be really listed here. If there is a field you want to return as a result of `defn`, then you can wrap it with the original struct on the way out:

    {defn_struct, ....} = call_defn(my_struct, ...)
    Map.merge(my_struct, Map.take(defn_struct, container_fields))

PR #170, opened 2021-12-01
Ctcloss

floatn:


seanmor5:
@floatn Sorry for delaying in the review process, I haven't been able to find time to read through the CTC Loss description and some example implementations to compare notes.

While you're waiting though, do you mind adding some tests so we can verify it works similar to PyTorch or tensorflow?

floatn:
I've compared results with pyTorch using this:
```python
!/usr/bin/python
import torch
import numpy as np

T = 10
C = 5
N = 1
S = 4

#input = torch.full(size=(T, N, C), fill_value = 1/C).log_softmax(2).detach().requires_grad_()
#input = torch.randn(T, N, C).log_softmax(2).detach().requires_grad_()
input = torch.tensor(
  [[[-0.9714, -0.9640, -2.9779, -3.0708, -1.9462]],
   [[-3.0251, -2.2230, -0.4155, -2.4767, -2.3113]],
   [[-1.5946, -1.2221, -2.1956, -2.1702, -1.2839]],
   [[-0.2618, -2.5582, -2.9246, -3.6593, -2.6112]],
   [[-2.2140, -0.4346, -3.3434, -3.2254, -1.7830]],
   [[-2.9288, -2.9082, -2.0663, -0.7128, -1.2909]],
   [[-3.3950, -1.9521, -0.2629, -4.4295, -3.1303]],
   [[-0.9938, -1.1471, -2.5672, -2.3438, -1.9693]],
   [[-2.1427, -3.1549, -3.6807, -1.1077, -0.7246]],
   [[-2.2734, -2.6316, -3.7343, -0.4695, -1.7381]]])
#target = torch.randint(low=1, high=C, size=(N, S), dtype=torch.long)
target = torch.tensor([[2, 3, 4, 1]])
input_lengths = torch.full(size=(N,), fill_value=T, dtype=torch.long)
target_lengths = torch.randint(low=S, high=S + 1, size=(N,), dtype=torch.long)
ctc_loss = torch.nn.CTCLoss(reduction='none')
loss = ctc_loss(input, target, input_lengths, target_lengths)
print(input)
print(target)
print(loss)
```

My concern is if it does suffice with gradient computation. Havn't investigated in that part yet.

In case this may be useful:
[https://ogunlao.github.io/blog/2020/07/17/breaking-down-ctc-loss.html](url)
[https://www.youtube.com/watch?v=e0ia-mN-7Kk](url)
original: [https://www.cs.toronto.edu/~graves/icml_2006.pdf](url)

lib/axon/losses.ex
```diff
@@ -1024,4 +1024,170 @@ defmodule Axon.Losses do
...
+
+  @doc """
+  WIP
+  y_pred should be log_softmax output of shape (B, T, D),
+  y_true should be of shape (B, S),
+  where
+  B - batch size,
+  T - sequence length,
+  D - dictionary size, including blank
+  S - maximum target length
```
seanmor5, line 1036: To be consistent with other functions in this module, this should go under an `Argument Shapes` section

lib/axon/losses.ex
```diff
@@ -1024,4 +1024,170 @@ defmodule Axon.Losses do
...
+  y_true nas to be preencoded, its first and last characters should be
+  blank, each initial character should be separated with blanks,
+  double blanks mark the end of target sequence.
+  """
+  defn ctcloss(y_true, y_pred, opts \\ []) do
+    opts = keyword!(opts, blank: 0, reduction: :none)
+    # eps = Nx.tensor([1.0e-7])
```
seanmor5, line 1043: Is this meant to be used?

lib/axon/losses.ex
```diff
@@ -1024,4 +1024,170 @@ defmodule Axon.Losses do
...
+          end
+
+        loss_b = -Nx.log(loss_b)
+        {Nx.put_slice(loss, [b], Nx.reshape(loss_b, {1})), b + 1, y_true, y_pred}
+      end
+
+    transform(
+      {opts[:reduction], loss},
+      fn
+        # {:mean, loss} -> Nx.mean(loss)
```
seanmor5, line 1067: Should this be uncommented?

lib/axon/losses.ex
```diff
@@ -1024,4 +1024,170 @@ defmodule Axon.Losses do
...
+  S - maximum target length
+  y_true nas to be preencoded, its first and last characters should be
+  blank, each initial character should be separated with blanks,
+  double blanks mark the end of target sequence.
+  """
+  defn ctcloss(y_true, y_pred, opts \\ []) do
+    opts = keyword!(opts, blank: 0, reduction: :none)
+    # eps = Nx.tensor([1.0e-7])
+    b_size = elem(Nx.shape(y_true), 0)
+    t_max = elem(Nx.shape(y_pred), 1) - 1
```
seanmor5, line 1045: We should do some shape checks to ensure input batch sizes match

lib/axon/losses.ex
```diff
@@ -1024,4 +1024,170 @@ defmodule Axon.Losses do
...
+    )
+  end
+
+  defn get_limits(y_true, t_max, blank) do
+    y_true_bb = Nx.concatenate([y_true, Nx.reshape(blank, {1})])
+    # Get `trimmed` tagret length.
+    {s_max, _} =
+      while {s = 0, y_true_bb}, y_true_bb[s] != y_true_bb[s + 1] do
+        {s + 1, y_true_bb}
+      end
```
seanmor5, line 1080: If I understand this right, we iterate over the tensor and select the first place where two adjacent values are the same? Off the top of my head, would this work? Assuming tokens are non-negative integers

```suggestion
s_max =
  y_true_bb
  |> Nx.window_reduce(-1, {2}, strides: [1], fn x, acc ->
    if acc == -1 do
      x # if I am seeing first value, just take it
    else
      x == acc # compare x2 and x1
    end
  end)
  |> Nx.multiply(Nx.iota({1, s})) # where s is the max targets?
  |> Nx.reduce_min() # take minimum index
```

The reason for avoiding while is that it is always slow

lib/axon/losses.ex
```diff
@@ -1024,4 +1024,170 @@ defmodule Axon.Losses do
...
+    {st_max, _, t_fin, _} =
+      while {st_max, s = 1, t = 1, y_true}, t <= t_max and s <= s_max - 2 do
+        s =
+          cond do
+            y_true[s] != y_true[s + 2] -> s + 2
+            true -> s + 1
+          end
+
+        {Nx.put_slice(st_max, [t], Nx.reshape(s, {1})), s, t + 1, y_true}
+      end
```
seanmor5, line 1093: Is it possible this could be a windowed operation as well?

lib/axon/losses.ex
```diff
@@ -1024,4 +1024,170 @@ defmodule Axon.Losses do
...
+          st_max
+
+        true ->
+          st_min = Nx.broadcast(0, {t_max + 1})
+
+          {st_min, _, _} =
+            while {st_min, dt = 1, st_max}, dt <= t_fin do
+              {Nx.put_slice(st_min, [t_max - dt + 1], Nx.reshape(st_max[t_fin - dt], {1})),
+               dt + 1, st_max}
+            end
```
seanmor5, line 1115: Does this need to be sequential?

lib/axon/losses.ex
```diff
@@ -1024,4 +1024,170 @@ defmodule Axon.Losses do
...
+  where
+  B - batch size,
+  T - sequence length,
+  D - dictionary size, including blank
+  S - maximum target length
+  y_true nas to be preencoded, its first and last characters should be
+  blank, each initial character should be separated with blanks,
+  double blanks mark the end of target sequence.
+  """
+  defn ctcloss(y_true, y_pred, opts \\ []) do
```
seanmor5, line 1041: To match the convention in this module, let's call it `connectionist_temporal_classification`. I know it's long, but all of the other implementations use the long form as well.

seanmor5 (commented):
Nice work! For now, I've left a few comments, mainly about rethinking some of the while-loops. If the operations must be sequential then that is fine, but I like to avoid while loops as much as possible as they are a big performance bottleneck

lib/axon/losses.ex
floatn: I was planning to rewrite it after review. Description is WIP.

lib/axon/losses.ex
floatn: Log(0) could be a thing. What do you think? Should we catch it or should we eps it?

lib/axon/losses.ex
floatn: "Mean" reducement mode as it is done in pyTorch, for example, will require some work. There it is calculated as an average over each target true length, and then over the batch. Do we need it such way or simple way?

lib/axon/losses.ex
floatn: Will do.

lib/axon/losses.ex
floatn: I'll check this.

lib/axon/losses.ex
floatn: I think it does. It's recursive by nature.

lib/axon/losses.ex
floatn: If you say so. But it's ctc_loss and CTCLoss in tf and torch.

lib/axon/losses.ex
seanmor5: I would use eps in that case

lib/axon/losses.ex
floatn: I can't use `Nx.put_slice` in one step, because I can't use `t_fin` in slice length, as long as it's scalar type, I think.

floatn: I think I know how to get rid of it. Will try.

seanmor5: Awesome! The more while loops we can get rid of the more performant this will be.

lib/axon/losses.ex
floatn: I have some issues with `window_reduce`.
This, for example, gives me a CompileError:
```elixir
defn test() do
  Nx.window_reduce(Nx.tensor([1, 2, 3]),
    0, {2}, [padding: :same],
    fn x, acc -> x + acc end
  )
end
```

lib/axon/losses.ex
floatn: Done.
It'd be nice to assert that maximum `y_true` value is lower that `y_pred` dictionary dim size, but that'll require some scalar vs scalar tensor action. Not quite sure how to do this.

lib/axon/losses.ex
seanmor5: What's the error?

floatn: <details>
<summary>Error log</summary>

```shell
== Compilation error in file lib/axon_ctc.ex ==
** (MatchError) no match of right hand side value: 1
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:623: Nx.BinaryBackend.element_wise_bin_op/4
    (nx 0.1.0-dev) lib/nx/defn/evaluator.ex:147: Nx.Defn.Evaluator.eval_apply/4
    (nx 0.1.0-dev) lib/nx/defn/evaluator.ex:63: Nx.Defn.Evaluator.eval/3
    (nx 0.1.0-dev) lib/nx/defn/evaluator.ex:83: anonymous fn/4 in Nx.Defn.Evaluator.eval_apply/4
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:1353: Nx.BinaryBackend."-window_reduce/6-_@5/2-2-"/4
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:1351: anonymous fn/9 in Nx.BinaryBackend.window_reduce/6
    (elixir 1.12.2) lib/enum.ex:2385: Enum."-reduce/3-lists^foldl/2-0-"/3
    (nx 0.1.0-dev) lib/nx/binary_backend.ex:1345: Nx.BinaryBackend.window_reduce/6
```
</details>

seanmor5: This seems to be a bug, can you extract a minimal example and post an issue in the Nx repo?

lib/axon/losses.ex
floatn: Didn't work out. Sorry.

lib/axon/losses.ex
floatn: I suggest we leave it as it is for now, and I'll come back to this when `window_reduce` bug will be settled.

floatn: Done.

lib/axon/losses.ex
floatn: I don't think it's possible.

floatn:
@seanmor5 Please review the changes.

lib/axon/losses.ex
seanmor5: Can you uncomment and just raise if `:reduction` is `:mean` for now?

seanmor5 (commented):
Just one comment

lib/axon/losses.ex
floatn: I'd better try to fully implement it now. Regarding of this, what if `y_true` would be a tuple also containing target lengths?

seanmor5: That's okay, as long as it's documented. I believe another loss function has that as well

lib/axon/losses.ex
```diff
@@ -1024,4 +1024,175 @@ defmodule Axon.Losses do
...
+
+  * `:reduction` - reduction mode. One of `:sum` or `:none`.
+    Defaults to `:none`.
+
+  ## Description
+    `l_true` contains lengths of target sequences. Nonzero positive values.
+    `y_true` contains target sequences. Each value represents a class
+    of element in range of available classes 0 <= y < D. Blank element
+    class is included in this range, but shouldn't be presented among
+    y_true values. Maximim target sequence length should be lower or equal
```
seanmor5, line 1047:
```suggestion
    y_true values. Maximum target sequence length should be lower or equal
```

lib/axon/losses.ex
```diff
@@ -1024,4 +1024,175 @@ defmodule Axon.Losses do
...
+    to `y_pred` sequence length: S <= T.
+    `y_pred` - logarithmized probabilities of classes D along the
```
seanmor5, line 1049: I think it's okay to just call this log probabilities since that's pretty common verbage

```suggestion
    `y_pred` - log probabilities of classes D along the
```

seanmor5 (approved):
Thank you for the patience! I have just made 2 suggestions on typos / wording in the docs (they're very small). This is great!

lib/axon/losses.ex
floatn: It's how they like to call it in pytorch docs.

PR #174, opened 2021-12-05
Start to test under JIT and add EXLA to test workflow

seanmor5:
WIP, still altering tests to JIT everything


josevalim:
In Nx master, you can do `Nx.Defn.default_compiler(EXLA)` inside the `setup` block. I think if you do this, you don't even need to rewrite the tests, as long as you call Nx.Defn.jit/3 from your anonymous functions?

josevalim:
To be clear, `init_fn.()` should for example run on EXLA by default if you set it as the default compiler.

josevalim:
Yeah, so this:

```elixir
  setup do
    Nx.Defn.default_options(compiler: Borked)
    :ok
  end
```

Make all tests fail, which means setting it to EXLA should be all!

lib/axon/compiler.ex
```diff
@@ -170,6 +172,8 @@ defmodule Axon.Compiler do
           expr
       end
     end
+
+    &Nx.Defn.jit_or_apply(predict_fn, [&1, &2])
```
seanmor5, line 176: @josevalim So previously I was not jitting anything when these functions we're called, but I think this is probably the better way to go as I think this will still work in a nested JIT (e.g. if you wanted to override the default compiler options) as well as within defn.

test/test_helper.exs
```diff
@@ -1,6 +1,11 @@
 ExUnit.start()
 
 defmodule AxonTestUtil do
+  def test_compiler do
+    use_exla? = System.get_env("USE_EXLA")
+    if use_exla?, do: EXLA, else: Nx.Defn.Evaluator
+  end
```
josevalim, line 7: Probably don't need this anymore?

lib/axon/compiler.ex
josevalim: Agreed! :+1:

josevalim (approved):

PR #176, opened 2021-12-08
Add default log handler

seanmor5:
This PR introduces an out of the box event handler for logging. The main purpose is to eliminate boilerplate when writing training loops (see removed code from examples). In the future, I see this method being extended to integrate nicely with dashboards such as Tensorboard and others.


lib/axon/loop.ex
```diff
@@ -695,6 +721,37 @@ defmodule Axon.Loop do
...
+
+  `message_fn` should take the loop state and return a binary
+  representing the message to be written to the IO device.
+  """
+  def log(%Loop{} = loop, event, message_fn, device \\ :stdio, filter \\ :always)
+      when is_function(message_fn, 1) do
+    log_fn = fn %State{} = state ->
+      # TODO: Should we rescue here and terminate the loop
+      # with an error gracefully, or just let the entire thing
+      # crash?
```
josevalim, line 745: I would go with a try/catch. If you use hooks, I think they are automatically try/catch-ed for you.

lib/axon/loop.ex
```diff
@@ -504,11 +504,37 @@ defmodule Axon.Loop do
...
       |> Axon.Loop.trainer(loss_weights)
       |> Axon.Loop.run(data)
+
+  ## Options
+
+    * `:log` - training loss and metric log interval. Set to 0 to silence
+      training logs. Defaults to 50
   """
-  def trainer(model, loss, optimizer) do
+  def trainer(model, loss, optimizer, opts \\ []) do
```
josevalim, line 513: Another option is to have a `|> Axon.Loop.attach_default_log_handler`.

lib/axon/loop.ex
seanmor5: I think this method specifically will not use hooks, but when we add training loop fusion then it will be necessary to use hooks and it will be necessary for this to rescue gracefully in some way

lib/axon/loop.ex
seanmor5: Do you mean rather than adding logging every time, make it explicit:

```elixir
model
|> Axon.Loop.trainer(:mean_squared_error, :sgd)
|> Axon.Loop.metric(:mean_absolute_error)
|> Axon.Loop.attach_default_log_handler()
|> Axon.Loop.run(data)
```

josevalim: Yes. It is just a suggestion though. :)

josevalim: Actually, I prefer your current approach. Ship it!

josevalim: Most will want something like that by default and it is like 10 chars to opt out, so current approach is good!

PR #178, opened 2021-12-10
Update Scidata version and callbacks examples

t-rutten:
Scidata `download/0` API [has changed slightly](https://github.com/elixir-nx/scidata/pull/18) and no longer accepts transform callbacks, so we can apply transformations to inputs and outputs after downloading.

### DIFF
--- a/examples/structured/credit_card_fraud.exs
+++ b/examples/structured/credit_card_fraud.exs
@@ -122,7 +122,6 @@ defmodule CreditCardFraud do
     model
     |> Axon.Loop.trainer(loss, optimizer)
     |> metrics()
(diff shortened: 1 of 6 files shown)


examples/structured/credit_card_fraud.exs
```diff
@@ -122,7 +122,6 @@ defmodule CreditCardFraud do
     model
     |> Axon.Loop.trainer(loss, optimizer)
     |> metrics()
-    |> Axon.Loop.handle(:iteration_completed, &log_metrics(&1, :train), every: 10)
```
t-rutten, line 125: Left over from #176

seanmor5:
Thanks!

PR #179, opened 2021-12-10
Add validation handler

seanmor5:
WIP to add a validation API to Axon loops. Validation sets are used as a "holdout" set to monitor and prevent overfitting. There are a few unresolved issues as of now:

1) How do we or do we even need to forward the outer loops compilation options to this inner loop?
2) Should we forward the model as well so it does not need to be explicitly specified? This would require adding some sort of `metadata` field to loop state.
3) How do we initialize state such that we are not changing the form of it and recompiling our train step after the first epoch because state now includes additional fields? I think this would probably go somewhat along with 1 and some sort of metadata field.

I'm also exploring different forms of this function to offer maximum control over the process. Overall though I think the base of how this works (e.g. computing the same metrics present in the loop and forwarding those for use in later handlers) is the way to go and I think it will make it easy to implement many other handlers.

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -512,29 +512,38 @@ defmodule Axon.Loop do
   """
   def trainer(model, loss, optimizer, opts \\ []) do
     log_interval = opts[:log] || 50
-    {init_fn, step_fn} = train_step(model, loss, optimizer)
+    # Build loss now so we can use it as a metric
+    loss_fn = build_loss_fn(loss)
+    {init_fn, step_fn} = train_step(model, loss_fn, optimizer)
     output_transform = fn state -> state.step_state[:model_state] end
-    loop = loop(step_fn, init_fn, output_transform)
+
+    loop =
+      step_fn
+      |> loop(init_fn, output_transform)
+      |> metric(loss_fn, "loss")
 
     if log_interval > 0 do
       loop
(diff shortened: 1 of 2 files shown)


seanmor5:
I have thought about it and addressed all of my main concerns. As a first draft implementation, I think this is good and we can iterate as problems arise. To answer the questions posed:

1) Inside the loop `run` method, prefer to use the compiler options set by `Nx.Defn.default_options` and only override if compiler is set. This means we can set compiler options at the top level and use those within validation, training, testing, etc.

2) In my mind at this time there's no reason to forward the model within loop state. In the future it may make sense in which case we can adjust, but as the model will already be readily available when creating the loop it's not a problem. It may be necessary to add `metadata` to the loop state later on though in which case we can reconsider.

3) This will be tracked in a separate issue to track how state changes in handlers before the loop is executed so we don't recompile unnecessarily

josevalim:
> Inside the loop run method, prefer to use the compiler options set by Nx.Defn.default_options and only override if compiler is set. This means we can set compiler options at the top level and use those within validation, training, testing, etc.

This gets my vote! The whole goal of `Nx.Defn.default_options` is to support projects like Axon and Meow, so you don't need to be carrying the compiler flag around. And if the compiler flag is given, you should just forward it to jit!

PR #182, opened 2021-12-21
Splitting uneven layers

stefkohub:
Hi all,
It seems in some cases it is useful to manage a bit more complex split layer scenario. 
The new functions are adding the possibility to use a list of sizes to split in uneven parts the tensors.
For instance:

```
iex(1)> Axon.input({1, 3, 416, 416})|>Axon.split([205, 211], axis: -1)
{----------------------------------------------------
                       Model
====================================================
 Layer                Shape              Parameters
====================================================
 input_1 ( input )    {1, 3, 416, 416}   0
 layer_2 ( custom )   {1, 3, 416, 205}   0
----------------------------------------------------
,
 ----------------------------------------------------
                       Model
====================================================
 Layer                Shape              Parameters
====================================================
 input_1 ( input )    {1, 3, 416, 416}   0
 layer_3 ( custom )   {1, 3, 416, 211}   0
----------------------------------------------------
}
```

What do you think?


seanmor5:
@stefkohub Nice! I like this, but I'm thinking of the name a little bit more, this is essentially a slice operation. I wonder if it is better to rename to something like that? Alternatively, split kind of makes sense because we are defining split sizes. cc @josevalim @polvalente for naming

lib/axon.ex
```diff
@@ -1528,10 +1528,34 @@ defmodule Axon do
...
+  In case `n` is a list the input graph is split in Enum.count slices with the 
+  size of the i-th element of the list.
   """
-  def split(%Axon{output_shape: shape} = parent, n, opts \\ []) do
+
+  def split(parent, n, opts \\ []) 
+
+  def split(%Axon{output_shape: shape} = parent, n, opts) when is_list(n) do
+    axis = opts[:axis] || -1
+    names = opts[:name] || []
```
seanmor5, line 1541: I wonder if for output names we default to something like `input_name_split_n` where `n` is the corresponding split and input name is the parent's name.

lib/axon.ex
```diff
@@ -1528,10 +1528,34 @@ defmodule Axon do
...
+    splits =
+      for {slice, i} <- n|>Enum.with_index do
+        layer(
+          parent,
+          fn x, _ -> Nx.slice_axis(x, i * slice, slice, axis) end,
+          Enum.at(split_shapes, i),
+          %{},
+          Enum.at(names, i)
+        )
+      end
```
seanmor5, line 1553:
```suggestion
    slice_names_shapes = Enum.zip([slice, split_shapes, names])
    splits =
      slice_names_shape
      |> Enum.with_index(fn {slice, shape, name}, i ->
        layer(parent, fn x, _ -> Nx.slice_axis(x, i * slice, slice, axis end, shape, %{}, name)
      end)
```

stefkohub:
Yes I think you are right on the name. For me it is totally OK to rename it. 

On the other hand, maybe we can consider also similar functions in similar frameworks such as Tensorflow, which is using a scalar or a tensor with any shape as split parameter. In that case they continue to call it "split".

lib/axon.ex
```diff
@@ -1528,10 +1528,34 @@ defmodule Axon do
   end
 
   @doc """
-  Splits input graph into a container of `n` input graphs
-  along the given axis.
+  Splits input graph into a container of `m` other input graphs
```
polvalente, line 1531: Does it work on all Axon graphs? If so, the usage of "input graph" here is kind of confusing.
If it's specific, I'd like to know more about what an input graph actually is and how it differs from other graphs

stefkohub: I performed some tests but I cannot answer for all the possible Axon graph.
Do you have in mind some specific test case? In so please describe it to me so I can test it.
Thank you.

lib/axon.ex
stefkohub: I was thinking to a default layer name but couldn't decide on something good. It is OK for me your proposal.

lib/axon.ex
polvalente: @stefkohub I asked because I'm not _that_ familiar with Axon, but I noticed that the whole docstring refers to "input graphs". So either it is correct and I'll learn something, or we should change it to something like "Splits the input graph into a container of `n` other graphs", and change accordingly below as well

seanmor5: In this case it is okay, `input graph` refers to any Axon object :)

stefkohub: Thank you @seanmor5 and @polvalente I am the first who needs to learn here :)

polvalente, line 1531:
```suggestion
  Splits input graph into a container of `m` other graphs 
```

@seanmor5 is this suggestion incorrect?

seanmor5: Ohhh yes this one you are correct!

lib/axon.ex
stefkohub: Something like this?

```
    names = opts[:name] || List.duplicate("#{parent.name}_split_", Enum.count(n))
    {_, split_shapes} = Axon.Shape.split(shape, n, axis)

    slice_names_shapes = Enum.zip([n, split_shapes, names])
    splits =
      slice_names_shapes
      |> Enum.with_index(fn {slice, shape, name}, i ->
        layer(parent, fn x, _ -> Nx.slice_axis(x, i * slice, slice, axis) end, shape, %{}, name<>"#{i}")
      end)
```

stefkohub:
Should I do something to allow the merging of this PR into the main branch?

seanmor5:
@stefkohub I think there are just 2 unresolved suggestions left and then we are good to go!

lib/axon.ex
stefkohub: How can I get the name of the parent layer? I noticed that now `parent.name` is returning a function expecting op and op_counts parameters, but I cannot understand what I must put into those two variables.

seanmor5:
@stefkohub In order to generate a name from the parent name function, you should have something like:

```
def split(%Axon{op: parent_op, name: parent_name_fn} = inp, ...) do
  child_name_fn = fn op, op_counts ->
    parent_name = parent_name_fn.(parent_op, op_counts)
    ...use parent name...
  end
```

You don't actually need to know the name at model building time :)

stefkohub:
nvm, I understood. I need to directly use `child_fn_name` in output Axon structure. Thank you again.

lib/axon.ex
```diff
@@ -1644,10 +1644,34 @@ defmodule Axon do
...
+  In case `n` is a list the input graph is split in Enum.count slices with the 
+  size of the i-th element of the list.
   """
-  def split(%Axon{output_shape: shape} = parent, n, opts \\ []) do
+
+  def split(parent, n, opts \\ []) 
+
+  def split(%Axon{output_shape: shape} = parent, n, opts) when is_list(n) do
+    axis = opts[:axis] || -1
+    names = opts[:name] || []
```
seanmor5, line 1657:
```suggestion
  def split(%Axon{name: parent_name_fn, output_shape: shape} = parent, n, opts) when is_list(n) do
    axis = opts[:axis] || -1
    names = opts[:name]
    names =
      case names do
        nil ->
          for i <- 0..n - 1 do
            fn op, op_counts -> parent_name_fn.(op, op_counts) <> "_i" end
          end
        
        names ->
          names
      end
```

This will give you names, `layer` should not be expecting a string in the most recent Axon :)

stefkohub: Thank you Sean for this suggestion. The problem is that at rows 120-125 of axon.ex into the current main branch we have:

```
  @doc type: :special
  def layer(parent, op, output_shape, parameters, name \\ nil, opts \\ [])
      when is_atom(op) or (is_function(op) and is_map(parameters)) do
    op_name = if is_atom(op), do: op, else: :layer

    {id, name} = unique_identifiers(op_name, name)
```
So it seems layer function is still expecting name as a string and not as a function. Should I add a check on name type into the layer function?

seanmor5:
@stefkohub I updated unique_identifiers so that it will also accept functions for name, so you should be unblocked

seanmor5:
@stefkohub I apologize this has gotten stale! There was a recent push to support transformers which added this functionality to Axon

PR #183, opened 2021-12-23
Fix: Nx.to_scalar in examples

floatn:
`Nx.to_scalar/1` has been renamed to `Nx.to_number/1`


seanmor5 (commented):
Thank you! A quick note - you can remove log_iteration from xor. That's from an earlier iteration and is no longer necessary

floatn:
@seanmor5 
Don't see such function in `xor.exs`. Did you mean `mnist_gan` example? It's used in a `Loop.log` there.

seanmor5:
It is the log_metrics function

floatn:
Oh. I see. `log_metrics` doesn't work atm. Don't we need logs in xor example? How about I just fix it?

PR #188, opened 2022-01-19
log function improved

msluszniak:
There is a problem with :io_lib.format in function supervised_log_message_fn/2. If a metric passed to the function is an integer type (like a number of true positive predictions), then :io_lib.format with options '~.5f' won't work correctly. To fix this I propose format_metric/1 which will handle this issue. I found the problem when I ran "credit_card_fraud.exs".

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -531,12 +531,20 @@ defmodule Axon.Loop do
     end
   end
 
+  defp format_metric({name, val}) do
+    {type, _} = val.type
+    case type do
+      :s -> "#{name}: #{:io_lib.format('~.B', [Nx.to_number(val)])}"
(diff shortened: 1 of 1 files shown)


lib/axon/loop.ex
```diff
@@ -531,12 +531,20 @@ defmodule Axon.Loop do
     end
   end
 
+  defp format_metric({name, val}) do
+    {type, _} = val.type
+    case type do
+      :s -> "#{name}: #{:io_lib.format('~.B', [Nx.to_number(val)])}"
+      :f -> "#{name}: #{:io_lib.format('~.7f', [Nx.to_number(val)])}"
+      _ -> "Unsupported type of metric"
```
seanmor5, line 539: We should also add formatting for u and bf types

seanmor5 (commented):
Good catch! Just one comment

PR #194, opened 2022-01-29
Lazily generate model names

seanmor5:
For Haskell enthusiasts

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -2208,7 +2208,7 @@ defmodule Axon do
         cache
         |> Enum.sort()
         |> Enum.unzip()
-        |> Kernel.elem(1)
+        |> elem(1)
         |> Enum.map(&Kernel.elem(&1, 0))
 
       rows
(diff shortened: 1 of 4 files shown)


lib/axon.ex
```diff
@@ -2202,13 +2202,14 @@ defmodule Axon do
...
       title = "Model"
       header = ["Layer", "Shape", "Parameters"]
-      {_, cache} = axon_to_rows(axon, %{})
+      {_, _, cache, _} = axon_to_rows(axon, %{}, %{})
 
       rows =
         cache
         |> Enum.sort()
         |> Enum.unzip()
         |> Kernel.elem(1)
```
polvalente, line 2211:
```suggestion
        |> elem(1)
```
Doesn't this work?

lib/axon.ex
```diff
@@ -2202,13 +2202,14 @@ defmodule Axon do
...
+        |> Enum.map(&Kernel.elem(&1, 0))
```
polvalente, line 2212:
```suggestion
        |> Enum.map(&elem(&1, 0))
```

PR #197, opened 2022-02-05
Add hooks

seanmor5:
WIP, need to add backward hooks

### DIFF
--- a/test/compiler_test.exs
+++ b/test/compiler_test.exs
@@ -1,6 +1,7 @@
 defmodule CompilerTest do
   use ExUnit.Case, async: true
   import ExUnit.CaptureLog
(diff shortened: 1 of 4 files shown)


test/compiler_test.exs
```diff
@@ -3857,4 +3858,37 @@ defmodule CompilerTest do
              }
     end
   end
+
+  describe "hooks" do
+    test "initialize hook" do
+      model =
+        Axon.input({nil, 1})
+        |> Axon.dense(1, kernel_initializer: :ones)
+        |> Axon.attach_hook(&IO.inspect/1, on: :initialize)
```
josevalim, line 3867: For testing, use something like `send(self(), :from_hook)` with `assert_received`. :)

PR #198, opened 2022-02-08
Linear resampling methods

floatn:


test/compiler_test.exs
```diff
@@ -2338,6 +2338,40 @@ defmodule CompilerTest do
       assert {init_fn, predict_fn} = Axon.compile(mp_model)
       assert Nx.type(predict_fn.(init_fn.(), Nx.random_uniform({1, 1, 3, 3}))) == {:bf, 16}
     end
+
+    test "bilinear without aligned corners" do
```
seanmor5, line 2342: These tests should be in layer_test.exs. The compiler tests test the Axon model to JIT compilation API

seanmor5 (approved):
This is awesome! Great work! Just the one comment about moving the tests and we are good to go!

test/compiler_test.exs
floatn: Done.

seanmor5 (approved):

PR #202, opened 2022-02-21
Do not send loop counters to device

josevalim:


polvalente (approved):

josevalim:
The CI failure is a precision error. The ONNX is also failing on main. :)

lib/axon/loop.ex
```diff
@@ -1065,10 +1061,7 @@ defmodule Axon.Loop do
               {:halt, {:halt_loop, state}}
 
             {:continue, state} ->
-              iters = Nx.to_number(iters)
-              max_iters = Nx.to_number(max_iters)
-
-              if iters > max_iters and max_iters != -1 do
+              if iters > max_iters and max_iters != nil do
```
josevalim, line 1064:
```suggestion
              if iters > max_iters and max_iters != -1 do
```

seanmor5 (approved):
Thank you :)

lib/axon/loop.ex
```diff
@@ -1052,23 +1044,32 @@ defmodule Axon.Loop do
...
+
+          {new_step_state, new_metrics} =
+            maybe_jit(batch_fn, [data, iters, step_state, metrics], jit_compile?, jit_opts)
+
+          state = %{
+            state
+            | iteration: iters + 1,
+              step_state: new_step_state,
+              metrics: new_metrics
+          }
```
josevalim, line 1062: @seanmor5 now it all works! I found out the simplest approach is actually to not pass the state into defn/jit. Only the parts that we need and then get them back. This way there is no confusion about the boundaries and about what goes and what comes back.

josevalim:
:green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #203, opened 2022-02-21
Improve horses or humans

josevalim:
This requires #202.

### DIFF
--- a/examples/vision/horses_or_humans.exs
+++ b/examples/vision/horses_or_humans.exs
@@ -5,7 +5,10 @@ Mix.install([
   {:nx, "~> 0.1.0-dev", github: "elixir-nx/nx", sparse: "nx", override: true}
 ])
 
-EXLA.set_preferred_defn_options([:tpu, :cuda, :rocm, :host])
+EXLA.set_preferred_defn_options(
+  [:tpu, :cuda, :rocm, :host],
+  run_options: [keep_on_device: true]
+)
 
 defmodule HorsesOrHumans do
   alias Axon.Loop.State
@@ -105,7 +108,7 @@ defmodule HorsesOrHumans do
     |> Axon.Loop.trainer(:binary_cross_entropy, optimizer)
     |> Axon.Loop.metric(:accuracy)
     |> Axon.Loop.handle(:iteration_completed, &log_metrics(&1, :train))
-    |> Axon.Loop.run(data, epochs: epochs, iterations: 100, compiler: EXLA, run_options: [keep_on_device: true])
+    |> Axon.Loop.run(data, epochs: epochs, iterations: 100)
   end
 
   def run() do
--- a/examples/vision/horses_or_humans.exs
+++ b/examples/vision/horses_or_humans.exs
@@ -1,12 +1,12 @@
 Mix.install([
-  {:flow, "~> 1.0"},
-  {:pixels, "~> 0.1.0"},
-  {:axon, "~> 0.1.0-dev", github: "elixir-nx/axon"},
+  {:stb_image, "~> 0.1.0"},
+  {:axon, "~> 0.1.0-dev", path: "."},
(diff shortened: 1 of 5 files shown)


examples/vision/horses_or_humans.exs
```diff
@@ -1,128 +1,128 @@
...
+    |> Stream.chunk_every(32, 32, :discard)
+    |> Task.async_stream(fn batch ->
+      {inp, labels} = batch |> Enum.map(&parse_png/1) |> Enum.unzip()
+      {Nx.stack(inp), Nx.stack(labels)}
+    end)
+    |> Stream.map(fn {:ok, {inp, labels}} -> {augment(inp), labels} end)
+    |> Stream.cycle()
+  end
+
+  # TODO: Fuse this with Axon network
```
josevalim, line 31:
```suggestion
```

examples/vision/horses_or_humans.exs
```diff
@@ -1,128 +1,128 @@
...
+    end
+  end
+
+  defp parse_png(filename) do
+    class =
+      if String.contains?(filename, "horses"),
+        do: Nx.tensor([1, 0], type: {:u, 8}),
+        else: Nx.tensor([0, 1], type: {:u, 8})
+
+    {:ok, binary, shape, :u8, :rgba} = StbImage.from_file(filename)
```
josevalim, line 49: Pixels does not work on mac/windows. :(

examples/vision/horses_or_humans.exs
```diff
@@ -1,128 +1,127 @@
...
+    |> Axon.max_pool(kernel_size: {2, 2})
+    |> Axon.conv(64, kernel_size: {3, 3}, activation: :relu)
+    |> Axon.max_pool(kernel_size: {2, 2})
+    |> Axon.flatten()
+    |> Axon.dropout(rate: 0.5)
+    |> Axon.dense(512, activation: :relu)
+    |> Axon.dense(2, activation: :softmax)
+  end
+
+  defp log_metrics(
```
josevalim, line 80: @seanmor5 I thought that we no longer needed to define this because we have a default callback? If so, how to use it? :)

examples/vision/horses_or_humans.exs
```diff
@@ -1,128 +1,129 @@
...
+
+  defnp augment(inp) do
+    # Normalize
+    inp = inp / 255.0
+
+    # For now just a random flip
+    if Nx.random_uniform({}) > 0.5 do
+      Nx.reverse(inp, axes: [-1])
+    else
+      Nx.reverse(inp, axes: [-2])
```
josevalim, line 41:
```suggestion
      Nx.reverse(inp, axes: [0])
    else
      Nx.reverse(inp, axes: [1])
```

examples/vision/horses_or_humans.exs
seanmor5: Yeah this can get deleted, logs will happen by default so there is no need!

examples/vision/horses_or_humans.exs
seanmor5: I'm fine with using StbImage anyway :)

examples/vision/horses_or_humans.exs
josevalim: I tried that but it didn’t log. I think i need to delete more stuff.

lib/axon/loop.ex
```diff
@@ -1069,6 +1064,8 @@ defmodule Axon.Loop do
               {:halt, {:halt_loop, state}}
 
             {:continue, state} ->
+              state = %{state | iteration: iters + 1}
```
josevalim, line 1067: This change is so we always log the first iteration (aka iteration 0). Otherwise the user may not have any feedback for a while.

josevalim: It also makes it consistent with epoch, where the epoch completed is called with epoch 0 and not the next epoch.

PR #204, opened 2022-02-22
Garbage collect tensors after each iteration

josevalim:


lib/axon/loop.ex
```diff
@@ -1054,6 +1054,9 @@ defmodule Axon.Loop do
           {new_step_state, new_metrics} =
             maybe_jit(batch_fn, [data, iters, step_state, metrics], jit_compile?, jit_opts)
 
+          # TODO: maybe deallocate data too based on an option?
```
josevalim, line 1057:
```suggestion
          # If the data is kept on device, it may take a while for the VM
          # to refcount and garbage collect them. So we immediately
          # deallocate. We may have the same issue with user data,
          # so we may want to provide an option for force deallocation.
          # TODO: deallocate data based on an option?
```

seanmor5:
I am curious if it is sufficient to do this after every epoch? Or perhaps in another process? I don't think it will be ~that~ expensive to do deallocations, but I'd be interested to see the performance impact

josevalim:
> I am curious if it is sufficient to do this after every epoch? 

I would say no. Given swapping from `backend_deallocate` to `garbage_collect` by itself on every iteration made a difference in using less memory, waiting for the epoch will probably use too much memory.

Issue #205, opened 2022-02-23
Add Axon.deserialize_params / Axon.serialize_params


josevalim:
We cannot serialize the result of training directly because it is Axon state (a custom struct), so we need to provide functions for wrapping/unwrapping it.

However, we have an issue in that if we wrap the `Nx.serialize/deserialize` functionality, the current Nx.serialize/deserialize does not allow us to attach any additional metadata to the serialized values, and we likely want to add at least an internal Axon version too.

One option is to allow a `:tag` to be given to `Nx.serialize`, which defaults to `:nx`. On deserialize, we return `{tag, value}`, which by default is `{:nx, container}`. This way we can make the `tag` be a custom map in Axon case, and on deserialization we could compare versions whatever else we want.

We could also serialize models but that's even more complicated.

seanmor5:
Given that this is directly related to Axon Loops, my proposal is to name this something like: `Axon.Loop.checkpoint` or `Axon.Loop.serialize_checkpoint` / `Axon.Loop.deserialize_checkpoint`. It will serialize and deserialize enough information about the loop state to begin training from the checkpoint after loading. It's pretty common to provide full last epoch training checkpoints, so that's what I'm thinking the purpose of this function will be.

PR #206, opened 2022-03-02
Initial draft of model serialization

### LINKED ISSUE #26: Add model import/export API
Need ability to serialize models to/from external formats. Model serialization is serialization of the actual computation graph. We should also have the ability to save and load model parameters, but I believe part of that discussion needs to happen upstream with a common `Nx` tensor serialization format. See e.g. https://github.com/elixir-nx/nx/issues/354

seanmor5:
Resolves #26 

Trying to find a better way to test

Note that #205 is more related to the Loop API and serializing training state. Similar to how in PyTorch you can save an entire state dict to resume training.

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -2402,8 +2402,8 @@ defmodule Axon do
   using `Nx.serialize/2`. There is some additional metadata included
   such as current serialization version for compatibility.
 
-  Serialization `opts` are forwarded to Nx.serialize and
-  `:erlang.term_to_binary` for controlling compression options.
+  Serialization `opts` are forwarded to `Nx.serialize/2` and
+  `:erlang.term_to_binary/2` for controlling compression options.
 
   ## Examples
 
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -2399,7 +2399,7 @@ defmodule Axon do
   Model and parameters are serialized as a tuple, where the
   model is converted to a recursive map to ensure compatibility
   with future Axon versions and the parameters are serialized
-  using `Nx.serialize`. There is some additional metadata included
+  using `Nx.serialize/2`. There is some additional metadata included
   such as current serialization version for compatibility.
 
   Serialization `opts` are forwarded to Nx.serialize and
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -85,6 +85,7 @@ defmodule Axon do
   @type t :: %__MODULE__{}
 
   @doc false
+  # Note: when adding new fields, consider how they fit serialize/deserialize
   @derive {
     Nx.Container,
     containers: [],
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -2495,7 +2495,7 @@ defmodule Axon do
     >
   """
   def deserialize(serialized, opts \\ []) do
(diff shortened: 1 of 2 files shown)


lib/axon.ex
```diff
@@ -2387,6 +2390,130 @@ defmodule Axon do
...
+  Serialization `opts` are forwarded to Nx.serialize and
+  `:erlang.term_to_binary` for controlling compression options.
+
+  ## Examples
+
+      iex> model = Axon.input({nil, 2}) |> Axon.dense(1, kernel_initializer: :zeros, activation: :relu)
+      iex> params = Axon.init(model)
+      iex> serialized = Axon.serialize(model, params)
+      iex> {saved_model, saved_params} = Axon.deserialize(serialized)
+      iex> saved_model
```
polvalente, line 2414: Perhaps there could be a test in which you `assert_all_close` on each layer, similarly to the `saved_params` assertions below, having used an orthogonal layer initialization or any other non-constant tensor value.
This would not be a doctest, though.

lib/axon.ex
```diff
@@ -2387,6 +2390,130 @@ defmodule Axon do
     end
   end
 
+  ## Serialization
+
+  @doc """
+  Serializes a model and it's parameters for persisting
```
josevalim, line 2396:
```suggestion
  Serializes a model and its parameters for persisting
```

lib/axon.ex
```diff
@@ -2387,6 +2390,130 @@ defmodule Axon do
...
+  models to disk or elsewhere.
+
+  Model and parameters are serialized as a tuple, where the
+  model is converted to a recursive map to ensure compatibility
+  with future Axon versions and the parameters are serialized
+  using `Nx.serialize`. There is some additional metadata included
+  such as current serialization version for compatibility.
+
+  Serialization `opts` are forwarded to Nx.serialize and
+  `:erlang.term_to_binary` for controlling compression options.
```
josevalim, line 2406:
```suggestion
  Serialization `opts` are forwarded to `Nx.serialize/2` and
  `:erlang.term_to_binary/2` for controlling compression options.
```

lib/axon.ex
```diff
@@ -79,6 +79,9 @@ defmodule Axon do
   """
   alias __MODULE__, as: Axon
 
+  # Axon serialization version
+  @file_version 1
+
   @type t :: %__MODULE__{}
 
   @doc false
```
josevalim, line 87:
```suggestion
  @doc false
  # Note: when adding new fields, consider how they fit serialize/deserialize
```

lib/axon.ex
```diff
@@ -2387,6 +2390,130 @@ defmodule Axon do
...
+
+  @doc """
+  Deserializes serialized model and parameters into a `{model, params}`
+  tuple.
+
+  It is the opposite of `Axon.serialize/3`.
+
+  ## Examples
+
+    iex> model = Axon.input({nil, 2}) |> Axon.dense(1, kernel_initializer: :zeros, activation: :relu)
```
josevalim, line 2467: Nitpick: indent with four spaces (or remove this altogether and tell them to check the docs for `Axon.serialize/3`.

josevalim (approved):
Looks great, just a bunch of minor nitpicks!

Issue #201, opened 2022-02-21
Resume training with `Axon.Loop.from_state`


PhillippOhlandt:
Hey,

resuming from an existing loop state does not seem to work properly right now.

For this, I am working in the horses_or_humans example.

I am returning the entire state in the `train_model` function:

```elixir
  defp train_model(model, data, optimizer, epochs) do
    model
    |> Axon.Loop.trainer(:binary_cross_entropy, optimizer)
    |> Axon.Loop.metric(:accuracy)
    |> Axon.Loop.handle(:iteration_completed, &log_metrics(&1, :train))
    |> Map.put(:output_transform, fn state -> state end)
    |> Axon.Loop.run(data, epochs: epochs, iterations: 100, compiler: EXLA)
  end
```

So I can store it and try to resume from that state during another execution:

```elixir
def run_continue() do
    case File.read("./horses_or_humans_continue_model") do
      {:ok, file} ->
        {model, model_state} = :erlang.binary_to_term(file)

        optimizer = Axon.Optimizers.adam(1.0e-4)
        centralized_optimizer = Axon.Updates.compose(Axon.Updates.centralize(), optimizer)

        data = data()
        IO.write("\n\nTraining model with gradient centralization\n\n")
        result =
          model
          |> Axon.Loop.trainer(:binary_cross_entropy, centralized_optimizer)
          |> Axon.Loop.from_state(model_state)
          |> Axon.Loop.metric(:accuracy)
          |> Axon.Loop.handle(:iteration_completed, &log_metrics(&1, :train))
          |> Map.put(:output_transform, fn state -> state end)
          |> Axon.Loop.run(data, epochs: 5, iterations: 100, compiler: EXLA)

        binary = :erlang.term_to_binary({model, result})
        File.write!("./horses_or_humans_continue_model", binary)
      _ ->
        model = build_model({nil, 4, 300, 300})
        optimizer = Axon.Optimizers.adam(1.0e-4)
        centralized_optimizer = Axon.Updates.compose(Axon.Updates.centralize(), optimizer)
        data = data()
        IO.write("\n\nTraining model with gradient centralization\n\n")
        result = train_model(model, data, centralized_optimizer, 5)
        binary = :erlang.term_to_binary({model, result})
        File.write!("./horses_or_humans_continue_model", binary)
    end
  end
```

But there seems to be an issue with anonymous functions:

```
[...]

Issue #142, opened 2021-10-20
Add out of the box event handlers


seanmor5:
Each example duplicates a `log_metrics` function which should be packaged into the API for convenience. As a good starting point, we should add the following event handlers to the API as a convenience:

- [x] Metrics Logging (see `log_metrics` in examples)
- [x] Model Checkpoints
- [ ] Early Stopping (this requires some thought because early stopping would be stop a trainer, but it would need to run in an evaluator)

One question is whether or not these should be implemented as functions in the API or in a separate module and then called to in the same way as metrics.

Case 1 (part of the API):

```elixir
loop
|> Axon.Loop.metric(:accuracy)
|> Axon.Loop.log(:iteration_completed, every: 50)
|> Axon.Loop.checkpoint(:epoch_completed)
|> Axon.Loop.run(...)
```

Case 2 (separate module):

```elixir
log and checkpoint defined in Axon.Loop.Handlers

loop
|> Axon.Loop.metric(:accuracy)
|> Axon.Loop.handle(:iteration_completed, :log, every: 50)
|> Axon.Loop.handle(:epoch_completed, :checkpoint)
|> Axon.Loop.run(...)
```

I lean towards case 1 because it provides a bit more readability when providing options to the handler rather than having to pass a parameterized function to `handle`. Case 2 keeps the API as barebones as possible though.

PR #208, opened 2022-03-03
First draft of loop checkpoints

### LINKED ISSUE #205: Add Axon.deserialize_params / Axon.serialize_params

seanmor5:
This is a draft of loop checkpoints.

This resolves #205 and should be sufficient to outline a resolution for #201.

It also addresses model checkpoint handler in #142.

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -363,7 +363,10 @@ defmodule Axon.Loop do
       }
     end
 
-    {init_fn, step_fn}
+    {
+      fn -> Nx.Defn.jit_or_apply(init_fn, []) end,
+      fn data, state -> Nx.Defn.jit_or_apply(step_fn, [data, state]) end
+    }
   end
 
   @doc """
@@ -854,8 +857,8 @@ defmodule Axon.Loop do
   By default, loop checkpoints will be saved at the end of every
   epoch in the current working directory under the `checkpoint/`
   path. Checkpoints are serialized representations of loop state
-  obtained from `Axon.Loop.serialize_state/3`. Serialization
-  options will be forwarded to `Axon.Loop.serialize_state/3`.
+  obtained from `Axon.Loop.serialize_state/2`. Serialization
+  options will be forwarded to `Axon.Loop.serialize_state/2`.
 
   You can customize checkpoint events by passing `:event` and `:filter`
   options:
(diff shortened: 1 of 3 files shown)


lib/axon/loop.ex
```diff
@@ -866,6 +987,49 @@ defmodule Axon.Loop do
...
+
+  It is the opposite of `Axon.Loop.serialize_state/3`.
+
+  By default, the step state is deserialized using `Nx.deserialize.2`;
+  however, this behavior can be changed if step state is an application
+  specific container. For example, if you introduce your own data
+  structure into step_state and you customized the serialization logic,
+  `Nx.deserialize/2` will not be sufficient for deserialization.
+  """
+  def deserialize_state(serialized, deserialize_step_state_fn \\ &Nx.deserialize/2, opts \\ []) do
```
josevalim, line 1025: I would make `deserialize_step_state_fn` part of opts to keep the API leaner?

lib/axon/loop.ex
```diff
@@ -866,6 +987,49 @@ defmodule Axon.Loop do
...
+    {1, state_map} = :erlang.binary_to_term(serialized, [:safe | opts])
+    step_state = deserialize_step_state_fn.(state_map.step_state, opts)
+    metrics = Nx.deserialize(state_map.metrics, opts)
+    state_map = %{state_map | step_state: step_state, metrics: metrics}
+    struct(Axon.Loop.State, state_map)
```
josevalim, line 1030:
```suggestion
    struct!(Axon.Loop.State, state_map)
```

So we raise on field mismatch?

lib/axon/loop.ex
```diff
@@ -866,6 +987,49 @@ defmodule Axon.Loop do
...
+  By default, the step state is serialized using `Nx.serialize/2`;
+  however, this behavior can be changed if step state is an application
+  specific container. For example, if you introduce your own data
+  structure into step_state, `Nx.serialize/3` will not be sufficient
+  for serialization.
+
+  `opts` controls serialization options such as compression. It is
+  forwarded to `:erlang.term_to_binary/3`.
+  """
+  def serialize_state(%State{} = state, serialize_step_state_fn \\ &Nx.serialize/2, opts \\ []) do
```
josevalim, line 1006: I would make `serialize_step_state_fn` part of opts to keep the API leaner?

lib/axon/loop.ex
```diff
@@ -846,6 +847,126 @@ defmodule Axon.Loop do
...
+
+      if save? do
+        filename = Path.join([path, file_pattern.(epoch, iter)])
+        dirname = Path.dirname(filename)
+        File.mkdir_p!(dirname)
+        File.write!(filename, serialized_state)
+        {:continue, updated_state}
+      else
+        {:continue, updated_state}
+      end
```
josevalim, line 962:
```suggestion
      end

      {:continue, updated_state}
```

josevalim (commented):
Just some minor nitpicks. We just need some tests and this is great!

PR #209, opened 2022-03-05
Add early stopping

### LINKED ISSUE #142: Add out of the box event handlers

seanmor5:
WIP

Resolves #142 

Needs tests and ability to stop on thresholds

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -1017,7 +1017,7 @@ defmodule Axon.Loop do
       {prev_criteria_value, since_last_improvement} =
         case handler_meta[:early_stop] do
           nil ->
-            nil
+            {nil, 0}
 
           meta ->
             {meta[monitor], meta[:since_last_improvement]}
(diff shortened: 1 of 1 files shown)


lib/axon/loop.ex
```diff
@@ -968,6 +968,95 @@ defmodule Axon.Loop do
...
+        raise ArgumentError,
+              "invalid metric to monitor, key #{inspect(monitor)} not present in metrics"
+      end
+
+      cur_criteria_value = metrics[monitor]
+
+      {prev_criteria_value, since_last_improvement} =
+        case handler_meta[:early_stop] do
+          nil ->
+            nil
```
floatn, line 1020: Will not this be a match error for `nil`?

josevalim (approved):

lib/axon/loop.ex
seanmor5: Good catch!

PR #226, opened 2022-03-28
cosine_similarity

vans163:
This needs the proper `@doc`

### DIFF
--- a/lib/axon/losses.ex
+++ b/lib/axon/losses.ex
@@ -883,6 +883,49 @@ defmodule Axon.Losses do
   end
 
   @doc ~S"""
+  Cosine Similarity error loss function.
+
+  $$l_i = \sum_i (\hat{y_i} - y_i)^2$$
+
+  ## Argument Shapes
+
+    * `y_true` - $(d_0, d_1, ..., d_n)$
+    * `y_pred` - $(d_0, d_1, ..., d_n)$
+
+  ## Options
+
+    * `:axes` - Defaults to `[1]`.
+    * `:eps` - Defaults to `0.000001`.
+
+  ## Examples
+      iex> y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
+      iex> y_true = Nx.tensor([[0.0, 1.0], [1.0, 1.0]])
+      #Nx.Tensor<
+        f32[2]
+        [0.0, 1.0]
+      >
+
+  ## Implemented in pytorch
+  https://github.com/pytorch/pytorch/blob/23902fb8955d48b8d1dfdd54c40e72c2744cac8e/aten/src/ATen/native/Distance.cpp#L242
+  """
+
+  defn cosine_similarity(y_true, y_pred, opts \\ []) do
+    opts = keyword!(opts, axes: [1], eps: 0.000001)
+
+    transform(
+      {opts[:axes], opts[:eps]},
+      fn
+        {axes, eps} ->
+          w12 = Nx.sum(y_true * y_pred, axes: axes)
+          w1 = Nx.sum(y_true * y_true, axes: axes)
+          w2 = Nx.sum(y_pred * y_pred, axes: axes)
+          n12 = Nx.sqrt(Nx.max(w1 * w2, eps * eps))
+          w12 / n12
+      end
+    )
+  end
+
+  @doc ~S"""
   Poisson loss function.
 
   $$l_i = \frac{1}{C} \sum_i^C y_i - (\hat{y_i} \cdot \log(y_i))$$
--- a/lib/axon/losses.ex
+++ b/lib/axon/losses.ex
@@ -906,7 +906,7 @@ defmodule Axon.Losses do
       iex> Axon.Losses.cosine_similarity(y_true, y_pred)
       #Nx.Tensor<
         f32[2]
-        [0.0, 1.0]
+        [0.0, 1.0000001192092896]
       >
   """
 
--- a/lib/axon/losses.ex
+++ b/lib/axon/losses.ex
@@ -894,33 +894,39 @@ defmodule Axon.Losses do
 
   ## Options
 
+    * `:reduction` - reduction mode. One of `:mean`, `:sum`, or `:none`.
+      Defaults to `:none`.
     * `:axes` - Defaults to `[1]`.
-    * `:eps` - Defaults to `0.000001`.
+    * `:eps` - Defaults to `1.0e-6`.
 
   ## Examples
+
       iex> y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
       iex> y_true = Nx.tensor([[0.0, 1.0], [1.0, 1.0]])
+      iex> Axon.Losses.cosine_similarity(y_true, y_pred)
       #Nx.Tensor<
         f32[2]
         [0.0, 1.0]
       >
-
-  ## Implemented in pytorch
-  https://github.com/pytorch/pytorch/blob/23902fb8955d48b8d1dfdd54c40e72c2744cac8e/aten/src/ATen/native/Distance.cpp#L242
   """
 
   defn cosine_similarity(y_true, y_pred, opts \\ []) do
-    opts = keyword!(opts, axes: [1], eps: 0.000001)
+    opts = keyword!(opts, axes: [1], eps: 1.0e-6, reduction: :none)
+    axes = opts[:axes]
+    eps = opts[:eps]
+
+    w12 = Nx.sum(y_true * y_pred, axes: axes)
+    w1 = Nx.LinAlg.norm(y_true, axes: axes)
+    w2 = Nx.LinAlg.norm(y_pred, axes: axes)
+    n12 = Nx.max(w1 * w2, eps)
+    loss = w12 / n12
 
     transform(
-      {opts[:axes], opts[:eps]},
+      {opts[:reduction], loss},
       fn
-        {axes, eps} ->
-          w12 = Nx.sum(y_true * y_pred, axes: axes)
-          w1 = Nx.sum(y_true * y_true, axes: axes)
-          w2 = Nx.sum(y_pred * y_pred, axes: axes)
-          n12 = Nx.sqrt(Nx.max(w1 * w2, eps * eps))
-          w12 / n12
+        {:mean, loss} -> Nx.mean(loss)
+        {:sum, loss} -> Nx.sum(loss)
+        {:none, loss} -> loss
       end
     )
   end
--- a/test/axon/losses_test.exs
+++ b/test/axon/losses_test.exs
@@ -231,4 +231,24 @@ defmodule Axon.LossesTest do
                Nx.tensor(2.472621440887451)
     end
   end
+
+  describe "cosine_similarity" do
+    test "supports eps" do
+      y_true = Nx.tensor([[0.0, 1.0], [1.0, 1.0]])
+      y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
+      eps = 1.0e-3
+
+      assert Axon.Losses.cosine_similarity(y_true, y_pred, eps: eps) ==
+               Nx.tensor([0.0, 1.0000001192092896])
+    end
+
+    test "supports axes" do
+      y_true = Nx.tensor([[0.0, 1.0], [1.0, 1.0]])
+      y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
+      axes = [0]
+
+      assert Axon.Losses.cosine_similarity(y_true, y_pred, axes: axes) ==
+               Nx.tensor([0.7071067690849304, 0.7071067690849304])
+    end
+  end
 end
--- a/test/axon/losses_test.exs
+++ b/test/axon/losses_test.exs
@@ -238,8 +238,10 @@ defmodule Axon.LossesTest do
       y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
       eps = 1.0e-3
 
-      assert Axon.Losses.cosine_similarity(y_true, y_pred, eps: eps) ==
-               Nx.tensor([0.0, 1.0000001192092896])
+      assert_all_close(
+        Axon.Losses.cosine_similarity(y_true, y_pred, eps: eps) ==
+          Nx.tensor([0.0, 1.0])
+      )
     end
 
     test "supports axes" do
@@ -247,8 +249,10 @@ defmodule Axon.LossesTest do
       y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
       axes = [0]
 
-      assert Axon.Losses.cosine_similarity(y_true, y_pred, axes: axes) ==
-               Nx.tensor([0.7071067690849304, 0.7071067690849304])
+      assert_all_close(
+        Axon.Losses.cosine_similarity(y_true, y_pred, axes: axes) ==
+          Nx.tensor([0.7071067690849304, 0.7071067690849304])
+      )
     end
   end
 end
--- a/test/axon/losses_test.exs
+++ b/test/axon/losses_test.exs
@@ -2,6 +2,8 @@ defmodule Axon.LossesTest do
   use ExUnit.Case, async: true
   doctest Axon.Losses
 
+  import AxonTestUtil
+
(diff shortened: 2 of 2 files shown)


lib/axon/losses.ex
```diff
@@ -958,6 +958,49 @@ defmodule Axon.Losses do
...
+
+  ## Examples
+      iex> y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
+      iex> y_true = Nx.tensor([[0.0, 1.0], [1.0, 1.0]])
+      #Nx.Tensor<
+        f32[2]
+        [0.0, 1.0]
+      >
+
+  ## Implemented in pytorch
```
polvalente, line 984:
```suggestion
  ## Implemented in PyTorch
```

lib/axon/losses.ex
```diff
@@ -958,6 +958,49 @@ defmodule Axon.Losses do
...
+      {opts[:axes], opts[:eps]},
+      fn
+        {axes, eps} ->
+          w12 = Nx.sum(y_true * y_pred, axes: axes)
+          w1 = Nx.sum(y_true * y_true, axes: axes)
+          w2 = Nx.sum(y_pred * y_pred, axes: axes)
+          n12 = Nx.sqrt(Nx.max(w1 * w2, eps * eps))
+          w12 / n12
+      end
+    )
```
polvalente, line 1001: Do we actually need this transform?

lib/axon/losses.ex
```diff
@@ -958,6 +958,49 @@ defmodule Axon.Losses do
...
+  defn cosine_similarity(y_true, y_pred, opts \\ []) do
+    opts = keyword!(opts, axes: [1], eps: 0.000001)
+
+    transform(
+      {opts[:axes], opts[:eps]},
+      fn
+        {axes, eps} ->
+          w12 = Nx.sum(y_true * y_pred, axes: axes)
+          w1 = Nx.sum(y_true * y_true, axes: axes)
+          w2 = Nx.sum(y_pred * y_pred, axes: axes)
```
polvalente, line 997:
```suggestion
          w1 = Nx.LinAlg.norm(y_true, axes: axes)
          w2 = Nx.LinAlg.norm(y_pred, axes: axes)
```

polvalente: And then you remove Nx.sqrt below

lib/axon/losses.ex
```diff
@@ -958,6 +958,49 @@ defmodule Axon.Losses do
...
+          n12 = Nx.sqrt(Nx.max(w1 * w2, eps * eps))
```
polvalente, line 998:
```suggestion
          n12 = Nx.max(w1 * w2, eps)
```

lib/axon/losses.ex
```diff
@@ -958,6 +958,49 @@ defmodule Axon.Losses do
...
+  """
+
+  defn cosine_similarity(y_true, y_pred, opts \\ []) do
+    opts = keyword!(opts, axes: [1], eps: 0.000001)
+
+    transform(
+      {opts[:axes], opts[:eps]},
+      fn
+        {axes, eps} ->
+          w12 = Nx.sum(y_true * y_pred, axes: axes)
```
polvalente, line 995: Shouldn't this be Nx.dot(y_true, y_pred)?

polvalente: Nevermind, this would only work for vectors

lib/axon/losses.ex
vans163: Im am not sure how to pass dynamic options then?

lib/axon/losses.ex
vans163: Perhaps its closer to the KERAS implementation then?
```elixir
  defn loss_cosine_similarity_keras(y_true, y_pred) do
    #https://github.com/keras-team/keras/blob/89879e2c76e86c685e44c47a6cdb82f7e645c142/keras/losses.py#L2167
    #y_true = tf.linalg.l2_normalize(y_true, axis=axis)
    #y_pred = tf.linalg.l2_normalize(y_pred, axis=axis)
    #return -tf.reduce_sum(y_true * y_pred, axis=axis)

    y_pred_l2 = Nx.divide(y_pred, Nx.LinAlg.norm(y_pred, axes: [1]) |> Nx.new_axis(1))
    y_true_l2 = Nx.divide(y_true, Nx.LinAlg.norm(y_true, axes: [1]) |> Nx.new_axis(1))
    -Nx.sum(y_true_l2 * y_pred_l2) / 2
  end
```
But im not sure to what degree these are equivalent (the PR returns a tensor, this one returns a number)

lib/axon/losses.ex
```diff
@@ -958,6 +958,49 @@ defmodule Axon.Losses do
...
+
+  ## Argument Shapes
+
+    * `y_true` - $(d_0, d_1, ..., d_n)$
+    * `y_pred` - $(d_0, d_1, ..., d_n)$
+
+  ## Options
+
+    * `:axes` - Defaults to `[1]`.
+    * `:eps` - Defaults to `0.000001`.
```
seanmor5, line 974:
```suggestion
    * `:eps` - Defaults to `1.0e-6`.
```

lib/axon/losses.ex
```diff
@@ -958,6 +958,49 @@ defmodule Axon.Losses do
...
+
+  ## Examples
+      iex> y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
```
seanmor5, line 977:
```suggestion
  ## Examples
 
      iex> y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
```

lib/axon/losses.ex
seanmor5: I would just remove this reference altogether
```suggestion
```

lib/axon/losses.ex
seanmor5: Options will work as normal if you just access them with `opts[:axes]` and `opts[:eps]`

lib/axon/losses.ex
vans163: this introduces a tiny rounding error with the example now returning instead of 1.0, not sure if this is significant.
```
#Nx.Tensor<
  f32[2] 
  [0.0, 1.0000001192092896]
>
```

seanmor5: That's small enough where I'm confident it won't be a huge deal

lib/axon/losses.ex
```diff
@@ -958,6 +958,55 @@ defmodule Axon.Losses do
...
+  defn cosine_similarity(y_true, y_pred, opts \\ []) do
+    opts = keyword!(opts, axes: [1], eps: 1.0e-6, reduction: :none)
+    axes = opts[:axes]
+    eps = opts[:eps]
+
+    w12 = Nx.sum(y_true * y_pred, axes: axes)
+    w1 = Nx.LinAlg.norm(y_true, axes: axes)
+    w2 = Nx.LinAlg.norm(y_pred, axes: axes)
+    n12 = Nx.max(w1 * w2, eps)
+    loss = w12 / n12
```
polvalente, line 998: @seanmor5 WDYT about moving the basic cosine similarity block to Scholar instead of keeping it here? We could actually do this with all of the loss functions here, in that case.

It would become something like:

```elixir
loss = Scholar.Metrics.cosine_similarity(y_true, y_pred, opts)
transform(...)
```

vans163:
There is a big problem here Nx.max is wrong, it should be Nx.clamp_min (does not exist), I am not sure how to clamp a tensor with Nx.

polvalente:
> There is a big problem here Nx.max is wrong, it should be Nx.clamp_min (does not exist), I am not sure how to clamp a tensor with Nx.

I think you can make use of Nx.clip for that:

```elixir
iex(2)> Nx.tensor([-1, 0, 1, 2, 3]) |> Nx.clip(1, 2)
#Nx.Tensor<
  s64[5]
  [1, 1, 1, 2, 2]
>
```

You'll also need to use Nx.Constants.max_finite(t.type) for the upper bound

lib/axon/losses.ex
```diff
@@ -958,6 +958,55 @@ defmodule Axon.Losses do
...
+
+  defn cosine_similarity(y_true, y_pred, opts \\ []) do
+    opts = keyword!(opts, axes: [1], eps: 1.0e-6, reduction: :none)
+    axes = opts[:axes]
+    eps = opts[:eps]
+
+    w12 = Nx.sum(y_true * y_pred, axes: axes)
+    w1 = Nx.LinAlg.norm(y_true, axes: axes)
+    w2 = Nx.LinAlg.norm(y_pred, axes: axes)
+    n12 = Nx.max(w1 * w2, eps)
```
polvalente, line 997: @vans163 bringing the discussion on "clip_min" here, what's the actual issue? My suggestion of Nx.clip actually does the same thing as Nx.max:

```elixir
iex(10)> t               
#Nx.Tensor<
  s64[5]
  [-1, 0, 1, 2, 3]
>
iex(11)> Nx.max(t, 1)
#Nx.Tensor<
  s64[5]
  [1, 1, 1, 2, 3]
>
```

josevalim:
AFAIK cosine_similarity is not in ONNX, so that may be a sign this indeed belongs to Scholar. :)

seanmor5:
@josevalim Most of the loss functions in Axon are not in ONNX either since they are mostly for training and ONNX is mostly focused on inference and converting models between frameworks, so I think it's okay to have this in Axon losses...it's in PyTorch and Keras/TensorFlow losses as well

josevalim:
I see, got ya!

josevalim:
I see! :+1: for this PR then.

In the future we may want to move this to Scholar too, so we will need to figure out if we will duplicate it, depend on Scholar, or create a separate library for this functionality. But we can postpone it for now.

PS: nice to see you around here @vans163!

lib/axon/losses.ex
seanmor5: @polvalente I am not yet ready to have Axon completely depend on Scholar, but in the future when Scholar gets built out more I think it makes sense

vans163:
I can rebase this once the code review is squared away. 

BTW I am using this like this. Maybe there should be an option to allow the subtraction, or would multiplying it by `-1` achieve the same thing?
```elixir
  defn loss_cosine_similarity(y_true, y_pred) do
    Nx.sum(1 - Axon.Losses.cosine_similarity(y_true, y_pred)) / Nx.axis_size(y_true, 0)
  end
```

vans163:
I am debugging this and the LinAlg version is much slower (or inaccurate) compared to the squareroot version on both the CPU and GPU, a huge factor slower the batch job for me does not even finish using LinAlg.

Workload is generating cosine similarity labels.

```elixir
defmodule MNISTCosine do
  import Nx.Defn

  @vec_num 10
  @max_sim 0.5
  @label_vec_dim 25

  defn gen_vec() do
    vec = Nx.random_normal({@label_vec_dim})
    Nx.power(Nx.power(vec, 2) / Nx.sum(Nx.power(vec, 2)), 0.5)
  end

  build_recursive = fn(build_recursive, idx)->
    nest = if idx == (@vec_num-1) do 0 else build_recursive.(build_recursive, idx+1) end
    str = """
    l = labels[#{idx}]
    sim = MNISTCosine.loss_cosine_similarity_torch(Nx.new_axis(new_vec,0), Nx.new_axis(l,0))
    if sim[0] <= #{@max_sim} do
      #{nest}
    else 1 end
    """
  end
  Code.compile_string("""
    defmodule Unroller do
      import Nx.Defn
      defn check_sim(found, labels, new_vec) do
        #{build_recursive.(build_recursive, 0)}
      end
    end
  """)

  defn add_label(labels, row, new_vec) do
    indices = Nx.stack([
      Nx.broadcast(row, {@label_vec_dim}),
      Nx.iota({@label_vec_dim})
    ], axis: -1)
    labels = Nx.indexed_add(labels, indices, new_vec)
  end

  defn generate_label_vectors_n() do
    labels = add_label(labels, 0, gen_vec())
    found = 0
    {_, labels} = while {found, labels}, Nx.less(found, @vec_num) do
      new_vec = gen_vec()
      is_similar = Unroller.check_sim(found, labels, new_vec)
      if not is_similar do
        labels = add_label(labels, found, new_vec)
        {found + 1, labels}
      else
        {found, labels}
      end
    end
  end

  def generate_label_vectors() do
    labels = [gen_vec()]
    Enum.reduce_while(1..100_000, labels, fn(_, labels)->
      v = gen_vec()
      is_similar = Enum.find(labels, fn(l)->
        sim = loss_cosine_similarity_torch(Nx.new_axis(v,0), Nx.new_axis(l,0))
        Nx.to_number(sim[0]) > @max_sim
      end)
      cond do
        length(labels) >= @vec_num -> {:halt, labels}
        !is_similar ->
          IO.inspect "found"
          {:cont, labels ++ [v]}
        true -> {:cont, labels}
      end
    end)
  end

  defn loss_cosine_similarity_torch(x1, x2) do
    dim = [1]
    eps = 0.000001

    w12 = Nx.sum(x1 * x2, axes: [1])
    w1 = Nx.sum(x1 * x1, axes: [1])
    w2 = Nx.sum(x2 * x2, axes: [1])
    n12 = Nx.sqrt(Nx.max((w1 * w2), eps*eps))

    #w12 = Nx.sum(x1 * x2, axes: dim)
    #w1 = Nx.LinAlg.norm(x1, axes: dim)
    #w2 = Nx.LinAlg.norm(x2, axes: dim)
    #n12 = Nx.max(w1 * w2, eps)

    w12 / n12
  end
end
```
MNISTCosine.generate_label_vectors_n()

seanmor5:
@vans163 Just checking in. So we are aware of issues with the RNG and are working for fixes upstream in Nx. Is there anything in particular that is blocking this PR from getting merged aside from my final review? From what I see from the code everything looks good :)

test/axon/losses_test.exs
```diff
@@ -231,4 +231,24 @@ defmodule Axon.LossesTest do
...
   end
+
+  describe "cosine_similarity" do
+    test "supports eps" do
+      y_true = Nx.tensor([[0.0, 1.0], [1.0, 1.0]])
+      y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
+      eps = 1.0e-3
+
+      assert Axon.Losses.cosine_similarity(y_true, y_pred, eps: eps) ==
+               Nx.tensor([0.0, 1.0000001192092896])
```
vans163, line 242: The precision can be reduced to `1.0000001`

test/axon/losses_test.exs
```diff
@@ -231,4 +231,24 @@ defmodule Axon.LossesTest do
...
+    end
+
+    test "supports axes" do
+      y_true = Nx.tensor([[0.0, 1.0], [1.0, 1.0]])
+      y_pred = Nx.tensor([[1.0, 0.0], [1.0, 1.0]])
+      axes = [0]
+
+      assert Axon.Losses.cosine_similarity(y_true, y_pred, axes: axes) ==
+               Nx.tensor([0.7071067690849304, 0.7071067690849304])
```
vans163, line 251: Precision can be reduced.

vans163:
@seanmor5 Have a look over I think this is final.

test/axon/losses_test.exs
seanmor5, line 242:
```suggestion
      assert_all_close(Axon.Losses.cosine_similarity(y_true, y_pred, eps: eps), Nx.tensor([0.0, 1.0]))
```

test/axon/losses_test.exs
seanmor5, line 251:
```suggestion
      assert_all_close(Axon.Losses.cosine_similarity(y_true, y_pred, axes: axes), Nx.tensor([0.7071067690849304, 0.7071067690849304]))
```

seanmor5 (commented):
Just those minor changes, to avoid precision issues!

lib/axon/losses.ex
```diff
@@ -906,7 +906,7 @@ defmodule Axon.Losses do
       iex> Axon.Losses.cosine_similarity(y_true, y_pred)
       #Nx.Tensor<
         f32[2]
-        [0.0, 1.0]
+        [0.0, 1.0000001192092896]
```
vans163, line 909: I dont get this one. `mix test` fails if its not like this.

PR #229, opened 2022-04-07
Support generic containers in Updates API

seanmor5:


lib/axon/shared.ex
```diff
@@ -87,59 +85,66 @@ defmodule Axon.Shared do
...
+      &deep_new(&1, fn x -> Axon.Initializers.full(value, shape: Nx.shape(x)) end)
     )
   end
 
   @doc """
   Deep merges two possibly nested maps, applying fun to leaf values.
   """
   def deep_merge(left, right, fun) do
-    Map.merge(left, right, &deep_resolve(&1, &2, &3, fun))
+    {merged, []} = Nx.Container.traverse(left, leaves(right), &recur_merge(&1, &2, fun))
```
josevalim, line 96: Perhaps we should have a case and when the second tuple element is not an empty list, we raise saying the shapes are not compatible.

lib/axon/updates.ex
```diff
@@ -45,7 +45,7 @@ defmodule Axon.Updates do
         transform(
           {x, step},
           fn {updates, step} ->
-            deep_new(updates, fn {k, x} -> Nx.multiply(x, step) end)
+            deep_new(updates, x -> Nx.multiply(x, step) end)
```
josevalim, line 48: SyntaxError :D

lib/axon/updates.ex
```diff
@@ -66,9 +66,9 @@ defmodule Axon.Updates do
       end
 
       defnp apply_my_update(updates, state) do
-        new_state = deep_new(state, fn {k, v} -> {k, Nx.add(v, 0.01)} end)
+        new_state = deep_new(state, v -> Nx.add(v, 0.01) end)
```
josevalim, line 69: SyntaxError :D

josevalim (approved):
Beautiful!

PR #230, opened 2022-04-12
Allow maps as inputs to predict

### LINKED ISSUE #218: Allow containers as input shapes
It would be nice to be able to specify some sort of container as an input to an Axon model. The current way to handle multiple inputs is to pass them all as a tuple:

```elixir
Axon.predict(model, params, {inp1, inp2})
```

But, some users may want to have something more structured such a custom struct or map which is an input to their model:

```elixir
Axon.predict(model, params, %{input_1: inp1, input_2: inp2})
```

seanmor5:
Resolves #218 

The initial way to work with multi-input models was to pass a tuple to `Axon.predict` where the order of inputs corresponded to the order inputs we're declared. So if you had a model:

```elixir
x = Axon.input({nil, 1})
y = Axon.input({nil, 1})
z = Axon.input({nil, 1})
model = Axon.add([x, y, z])
```

You would pass values for `x`, `y`, and `z` in order as:

```
Axon.predict(model, params, {x_val, y_val, z_val})
```

Because `x` was declared before `y` was declared before `z`. This was done before Nx supported maps, so it was the easiest way to accomplish multi-inputs. We sort inputs by the Layer's ID because it's a monotonically increasing random integer, and this was the easiest way to ensure inputs we're mapped correctly.

I think now because we support generic containers it makes sense to instead pass multiple inputs strictly with maps. In my opinion the code is more explicit and readable, and you don't have to go to the docs to figure out how to pass multiple inputs in when looking at a multi-input example:

```elixir
x = Axon.input({nil, 1}, name: :x)
y = Axon.input({nil, 1}, name: :y)
z = Axon.input({nil, 1}, name: :z)
model = Axon.add([x, y, z])

Axon.predict(model, params, %{x: x_val, y: y_val, z: z_val})
```

### DIFF
--- a/examples/basics/multi_input_example.exs
+++ b/examples/basics/multi_input_example.exs
@@ -8,8 +8,8 @@ defmodule XOR do
   require Axon
 
   defp build_model(input_shape1, input_shape2) do
-    inp1 = Axon.input(input_shape1)
-    inp2 = Axon.input(input_shape2)
+    inp1 = Axon.input(input_shape1, name: :x1)
+    inp2 = Axon.input(input_shape2, name: :x2)
     inp1
     |> Axon.concatenate(inp2)
     |> Axon.dense(8, activation: :tanh)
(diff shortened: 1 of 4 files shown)


josevalim:
Agreed, this is beauytiful, .ship it!

PR #231, opened 2022-04-15
Add Axon.container

### LINKED ISSUE #219: Add Axon.container
With the new container API as well as the new "lazy names" approach we take during compilation, I think it's necessary to add a new combinator:

```elixir
out1 = Axon.input({nil, 1})
out2 = Axon.input({nil, 1})

out = Axon.container({out1, out2})
```

And `container` accepts anything that implements the `Nx.Container` protocol. My biggest concern with this API is that we will use flexibility by wrapping containers in the Axon data structure. For example this would potentially mean you couldn't just pattern match on RNN results:

```elixir
{state, out} = Axon.lstm(...)
```

Because hypothetically the result would be wrapped in `Axon.container`

seanmor5:
Resolves #219 

I ended up refactoring the compiler a lot because I realized how I was doing things before was wrong, so it's much cleaner now IMO

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -217,6 +217,46 @@ defmodule Axon do
   end
 
   @doc """
+  Adds a container layer to the network.
+
+  In certain cases you may want your model to have mulitple
+  outputs. In order to make this work, you must "join" the
+  outputs into an Axon layer using this function for use in
+  initialization and inference later on.
+
+  The given container can be any valid Axon Nx container.
+
+  ## Options
+
+    * `:name` - Layer name.
+  """
+  @doc type: :special
+  def container(container, opts \\ []) do
+    output_shape =
+      deep_new(container, fn %Axon{output_shape: shape} ->
+        shape
+      end)
+
+    layer(container, :container, output_shape, %{}, opts[:name], opts)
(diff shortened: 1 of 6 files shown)


lib/axon.ex
```diff
@@ -216,6 +216,46 @@ defmodule Axon do
             " a number, wrap it with a call to Nx.tensor/2"
   end
 
+  @doc """
+  Adds a container layer to the network.
+
+  In certain cases you may want your model to have mulitple
```
josevalim, line 222:
```suggestion
  In certain cases you may want your model to have multiple
```

lib/axon.ex
```diff
@@ -216,6 +216,46 @@ defmodule Axon do
...
+  outputs. In order to make this work, you must "join" the
+  outputs into an Axon layer using this function for use in
+  initialization and inference later on.
+
+  The given container can be any valid Axon Nx container.
+
```
josevalim, line 228: Can you include an example of a simple network with multiple outputs?

josevalim (commented):
Beautiful diff! 400LOC less!

josevalim (approved):

PR #233, opened 2022-04-15
Make options optional in updates

### LINKED ISSUE #222: Make options optional in Axon.Updates
Right now all of the Axon.Updates functions require a list of options, so they're not really optional. I'm not a fan of that API, so it should be refactored to make the actual options optional

seanmor5:
Resolves #222 

While the cases are more verbose I think it makes the API much friendlier

### DIFF
--- a/lib/axon/updates.ex
+++ b/lib/axon/updates.ex
@@ -119,9 +119,7 @@ defmodule Axon.Updates do
     * [Adam: A Method for Stochastic Optimization](https://arxiv.org/abs/1412.6980)
 
   """
-  def scale_by_adam() do
-    scale_by_adam(identity(), [])
-  end
+  def scale_by_adam(combinator_or_opts \\ [])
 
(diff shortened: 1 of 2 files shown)


lib/axon/updates.ex
```diff
@@ -163,7 +177,21 @@ defmodule Axon.Updates do
       * `:eps` - numerical stability term. Defaults to `1.0e-7`
 
   """
-  def scale_by_rss(combinator \\ identity(), opts) do
+  def scale_by_rss() do
+    scale_by_rss(identity(), [])
+  end
+
+  def scale_by_rss(opts) when is_list(opts) do
```
josevalim, line 184:
```suggestion
  def scale_by_rss(combinator_or_opts \\ [])

  def scale_by_rss(opts) when is_list(opts) do
```

You can apply similar technique everywhere. :) It will save you 2 LOC per function :P

PR #237, opened 2022-05-01
Fix Default Value of eps_root in Adam Documentation

nskins:
Updates documentation to specify the true default value of `eps_root` used by the Adam optimizer.


seanmor5:
Thanks!

PR #246, opened 2022-05-06
Compiler fix

seanmor5:


lib/axon/compiler.ex
```diff
@@ -279,38 +304,43 @@ defmodule Axon.Compiler do
...
     op_counts = Map.update(op_counts, op, 1, fn x -> x + 1 end)
 
-    input =
-      res
-      |> safe_as_type(compute)
-      |> apply_hooks(:pre_forward, mode, hooks)
+    fun = fn params, inputs, state, cache ->
+      {res, state} =
+        params
+        |> cache[parent_id].(inputs, state, cache)
```
josevalim, line 317: This should be `call_cache`, no?

josevalim (approved):
One comment about a missing call_cache and ship it

lib/axon.ex
```diff
@@ -1810,12 +1810,32 @@ defmodule Axon do
       unrolling of RNN.
 
   """
-  @doc type: :recurrent
-  def lstm(%Axon{output_shape: shape} = x, units, opts \\ [])
+  def lstm(%Axon{} = x, units) when is_integer(units) and units > 0 do
```
josevalim, line 1813: I think you need to add `@doc type: :recurrent` to all clauses. And then we usually do:

```elixir
@doc """
See `lstm/4`.
"""
@doc type: :recurrent
def lstm(%Axon{} = x, units) do
```

I typically tend to document the one with the biggest arity, but any is fine, as long as it is consitent in the API. You may need to add more examples to Axon.lstm docs too.

lib/axon.ex
```diff
@@ -1941,23 +1965,40 @@ defmodule Axon do
       unrolling of RNN.
 
   """
+  def gru(%Axon{} = x, units) do
```
josevalim, line 1968: Same notes as lstm.

lib/axon.ex
```diff
@@ -2050,12 +2102,33 @@ defmodule Axon do
       unrolling of RNN.
 
   """
+  def conv_lstm(%Axon{} = x, units) when is_integer(units) and units > 0 do
```
josevalim, line 2105: Same notes as lstm.

PR #249, opened 2022-05-07
Handle atom names in Axon Inspect

alisinabh:
Hey :wave: 

In my tests I have realized passing atom names to some layers like `lstm` will result in errors. I don't know if not supporting atoms here were intentional but I have developed this tiny fix with some tests for that.

Let me know if it could be improved or feel free to close this if this is intended behavior.


alisinabh:
Closed in favor of #255

PR #255, opened 2022-05-10
Standardize layer names on binaries

### LINKED ISSUE #245: Standardize on layer names as atoms or strings
It's possible to switch up between atoms and strings as layer names. We should standardize on one. It's possible to run into situations especially with input maps where you use an atom with the correct names but it raises because it's expecting a string, and the error is not clear

alisinabh:
Closes #245


seanmor5 (approved):
Beautiful! Thank you!

PR #256, opened 2022-05-10
Unify argument names in normalization layers

jonatanklosko:
The layer docs and the top-level API use the word "beta".


seanmor5 (approved):
Thank you!

PR #258, opened 2022-05-11
Revert to last working compiler

seanmor5:


seanmor5:
So my error was a dumb one...I was adding the `graph` in at the top level in order to raise `Axon.CompilerError`

PR #259, opened 2022-05-11
Resnet50 livebook

msluszniak:


notebooks/resnet50.livemd
```diff
@@ -0,0 +1,152 @@
+# ResNet50
+
+## Introduction
+
+ResNet, which stands for Residual Network, is a CNN which won in 2015 ILSVRC, a renowned and prestige competition for Computer Vision scientists. The winning version of ResNet has 152 layers, but all concepts stay the same with 50 layers.
```
seanmor5, line 5:
```suggestion
ResNet, which stands for Residual Network, is a Convolutional Neural Network (CNN) which won ILSVRC 2015. ILSVRC is a renowned and prestigious competition for Computer Vision scientists. The winning version of ResNet has 152 layers, but there are popular variants with fewer layers.
```

notebooks/resnet50.livemd
```diff
@@ -0,0 +1,152 @@
...
+
+The key concept introduced in ResNet is a skip connection.
+
+<!-- <img src="https://upload.wikimedia.org/wikipedia/commons/5/5f/ResNets.svg"> -->
+
+| ![](https://upload.wikimedia.org/wikipedia/commons/5/5f/ResNets.svg) |
+| :------------------------------------------------------------------: |
+| Figure 1: An example of skip connection                              |
+
+As we can see in the picture above, an input to layer l is not only output from layer l-1 but also output from l-2. In other words, a signal from l-2 skips the l-1 in one of the connections.
```
seanmor5, line 15:
```suggestion
As we can see in the picture above, an input to layer l is not only output from layer l-1 but also output from l-2. In other words, a signal from l-2 skips l-1 in one of the connections.
```

notebooks/resnet50.livemd
```diff
@@ -0,0 +1,152 @@
...
+
+The ResNet architecture has a few benefits over NN without skip connections.
```
seanmor5, line 17:
```suggestion
The ResNet architecture has a few benefits over neural networks without skip connections.
```

seanmor5 (commented):
Can you add an example prediction at the end of the example?

notebooks/resnet50.livemd
```diff
@@ -0,0 +1,152 @@
...
+
+We can say that ResNet consists of Residual Units (RU). Even if one RU stops the learning process, this will not affect other RUs.
+
+## Imports
+
+The only package we need in this example is Axon.
+
+```elixir
+Mix.install([
+  {:axon, "~> 0.1.0-dev", github: "elixir-nx/axon"}
```
seanmor5, line 31:
```suggestion
  {:axon, "~> 0.1"}
```

seanmor5:
@msluszniak Given some work in the other library that will also have an implementation of ResNet, I think that is a better place for a demonstration of ResNet usage/implementation. Thank you for your hard work regardless :)

PR #260, opened 2022-05-12
Simplify compiler

seanmor5:
This PR aims to greatly simplify the compiler by tweaking a few things and following a few unifying principles. I think the goal is for this to be the final evolution of the layer API into an interface that is flexible, simple, and easy to understand. Overall here's what needs to happen/change:

1) Parameters need to be defined with `Axon.param` and passed as a list to `layer` in the order they are to be passed to their implementation function. The original usage of map I think was for order preservation/forwarding, but it's actually not necessary and creates some ambiguity about what the name of the parameter being referenced is. I think the list is more clear as the order you give parameters in the list is the order they will be used. Another option even is to just pass everything (inputs and params) as inputs, and let `Axon.layer` split parameters in inputs by struct type. This also allows us to unify pretty much every trainable layer clause because we don't hard match on parameters---we just consume the list and forward it to the implementation function.

2) We need to forward `:mode` as an option to every implementation function. This means every implementation function needs to have a trailing `opts`---even if they are not used. I think though this is favorable over being ambiguous. And this means custom layers can have behavior that depends on training and inference mode---just like dropout and batch norm. This was not possible before.

3) We need to come up with an API for "stateful" layers so the compiler knows how to handle those in training mode. This will allow us to eliminate the additional `:batch_norm` clause and will mean that custom layers can also implement layers with state.

4) We need a more graceful way to implement combinators such as `add`, `concatenate`, and `cond` as functions in `Axon.Layers` so that we can unify them under the current monolithic clause.

Once this is complete custom layers will be 100% consistent with Axon's internal layers and everything that's possible to do internally with Axon will be possible externally with custom layers. So the flow of custom layers will be:

1. Define an implementation function as `defn` which can be an arbitrary arity function, as long as it ends in `opts`.
2. Define an interface for the layer which declares trainable parameters and state and attaches to the implementation function.

I also think it will be really cool to just say the Axon compiler is basically 1 function

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -78,6 +78,7 @@ defmodule Axon do
         |> Axon.Loop.run(train_data, epochs: 10, compiler: EXLA)
   """
   alias __MODULE__, as: Axon
+  alias Axon.Parameter
 
   # Axon serialization version
   @file_version 1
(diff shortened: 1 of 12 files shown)


josevalim:
I like this direction a lot. Option 3 should also help us address the existing issue that we are not passing the `model_state` as parameters to JIT, which means they are embedded in the models and that causes other issues.

PR #265, opened 2022-05-30
Explicit backend transfer of constants

seanmor5:
This seems to be the tweak that allows EXLA Backend and Compiler to coexist within Axon. I'm still troubleshooting an issue that EXLA raises though about casting types (at least when performing inference on various transformer models)

### DIFF
--- a/lib/axon/compiler.ex
+++ b/lib/axon/compiler.ex
@@ -202,6 +202,8 @@ defmodule Axon.Compiler do
          {cache, op_counts},
          _
        ) do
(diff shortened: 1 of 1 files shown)


josevalim:
Shit. I see the issue. You have two options:

1. convert this to the binary backend and the binary backend will introduce it as a constant into the model.
2. put this as part of the model_state (the same one used by batch_norm) so you can actually pass it as arg to jit

Thoughts?

seanmor5:
Given that the constants are should be mostly small, I think transferring to BinaryBackend is probably best

PR #266, opened 2022-06-05
Fix accuracy calculation logic

alisinabh:
Using compare operators inside `defn` even with `transform` will not work as expected. 
To do that it needs to be inside a separate function.

### DIFF
--- a/lib/axon/metrics.ex
+++ b/lib/axon/metrics.ex
@@ -40,27 +40,37 @@ defmodule Axon.Metrics do
 
   ## Examples
 
+      iex> Axon.Metrics.accuracy(Nx.tensor([[1], [0], [0]]), Nx.tensor([[1], [1], [1]]))
+      #Nx.Tensor<
+        f32
+        0.3333333432674408
+      >
+
       iex> Axon.Metrics.accuracy(Nx.tensor([[0, 1], [1, 0], [1, 0]]), Nx.tensor([[0, 1], [1, 0], [0, 1]]))
       #Nx.Tensor<
         f32
         0.6666666865348816
       >
 
+      iex> Axon.Metrics.accuracy(Nx.tensor([[0, 1, 0], [1, 0, 0]]), Nx.tensor([[0, 1, 0], [0, 1, 0]]))
+      #Nx.Tensor<
+        f32
+        0.5
+      >
+
   """
   defn accuracy(y_true, y_pred) do
-    transform({y_true, y_pred}, fn {y_true, y_pred} ->
-      if elem(Nx.shape(y_pred), Nx.rank(y_pred) - 1) == 1 do
-        y_pred
-        |> Nx.greater(0.5)
-        |> Nx.equal(y_true)
-        |> Nx.mean()
-      else
-        y_true
-        |> Nx.argmax(axis: -1)
-        |> Nx.equal(Nx.argmax(y_pred, axis: -1))
-        |> Nx.mean()
-      end
-    end)
+    if elem(Nx.shape(y_pred), Nx.rank(y_pred) - 1) == 1 do
+      y_pred
+      |> Nx.greater(0.5)
+      |> Nx.equal(y_true)
+      |> Nx.mean()
+    else
+      y_true
+      |> Nx.argmax(axis: -1)
+      |> Nx.equal(Nx.argmax(y_pred, axis: -1))
+      |> Nx.mean()
+    end
   end
 
   @doc ~S"""


lib/axon/metrics.ex
```diff
@@ -46,21 +46,29 @@ defmodule Axon.Metrics do
...
-        |> Nx.argmax(axis: -1)
-        |> Nx.equal(Nx.argmax(y_pred, axis: -1))
-        |> Nx.mean()
-      end
-    end)
+    transform({y_true, y_pred}, &calculate_accuracy/1)
+  end
+
+  defp calculate_accuracy({y_true, y_pred}) do
+    if elem(Nx.shape(y_pred), Nx.rank(y_pred) - 1) == 1 do
```
polvalente, line 61: I think just this equality should be done inside transform, returning a "boolean" tensor. The rest of the calculation can be  done outside of transform

alisinabh: Good point. I have added f0a4d64263d294ca0f90e38c05abf2489ea0e253 fixup. Let me know if there is a better way to write this.

lib/axon/metrics.ex
```diff
@@ -46,21 +46,29 @@ defmodule Axon.Metrics do
...
+    else
+      y_true
+      |> Nx.argmax(axis: -1)
+      |> Nx.equal(Nx.argmax(y_pred, axis: -1))
+      |> Nx.mean()
+    end
+  end
+
+  defp binary_prediction?(y_pred) do
+    if elem(Nx.shape(y_pred), Nx.rank(y_pred) - 1) == 1, do: 1, else: 0
```
josevalim, line 71: Unfortunately the previous code was better, until we perform branch folding in the `defn` compiler.

```suggestion
    transform({y_true, y_pred}, &accuracy/2)
  end

  defp accuracy({y_true, y_pred}) do
    if elem(Nx.shape(y_pred), Nx.rank(y_pred) - 1) == 1 do
      y_pred
      |> Nx.greater(0.5)
      |> Nx.equal(y_true)
      |> Nx.mean()
    else
      y_true
      |> Nx.argmax(axis: -1)
      |> Nx.equal(Nx.argmax(y_pred, axis: -1))
      |> Nx.mean()
    end
```

josevalim: I have implemented branch elimination in this PR: https://github.com/elixir-nx/nx/pull/802

The good news is that we no longer even need the `transform` at all. :)

alisinabh: I have added 0c23719f5fcb0b67134608b8041337e0f2c10ce4 which removes the transform. The good news is it is also working without using the changes you made in elixir-nx/nx#802. I'm not sure how!

To see how it works I have used the following code which adds some debug messages.

```elixir
  defn accuracy(y_true, y_pred) do
    if elem(Nx.shape(y_pred), Nx.rank(y_pred) - 1) == 1 do
      transform(y_pred, fn _ -> IO.puts("BIN") end)

      y_pred
      |> Nx.greater(0.5)
      |> Nx.equal(y_true)
      |> Nx.mean()
    else
      transform(y_pred, fn _ -> IO.puts("CAT") end)

      y_true
      |> Nx.argmax(axis: -1)
      |> Nx.equal(Nx.argmax(y_pred, axis: -1))
      |> Nx.mean()
    end
  end
```

But what I get using either your PR or the hex package is this.
```elixir
iex(2)> Axon.Metrics.accuracy(Nx.tensor([[1], [0], [0]]), Nx.tensor([[1], [1], [1]]))
BIN 
CAT 
#Nx.Tensor<
  f32
  0.3333333432674408
>
```

Which to me indicates that it is running both branches. Is this the intended behaviour?

josevalim: Yes, both branches are always executed, because the Elixir code is executed at compilation time.

The issue with the previous code is that we would then send both branches to the CPU/GPU compiler. My patch makes it so the compiler is smart enough to send only one branch!

josevalim: I think the current version is good as is. I assume all compilers will be able to optimize it. The other concern could be `grad` but I don't think this function is grad-ed.

alisinabh: Great. Thanks for the explanation.
Squashed and ready :)

seanmor5:
Thank you @alisinabh ! Is there any way you can implement a small test for each branch? Just to validate it's working as expected?

alisinabh:
Hello @seanmor5,
I have added these two doctests which each one test a branch of this function.

```elixir
  iex> Axon.Metrics.accuracy(Nx.tensor([[1], [0], [0]]), Nx.tensor([[1], [1], [1]]))
  #Nx.Tensor<
    f32
    0.3333333432674408
  >


  iex> Axon.Metrics.accuracy(Nx.tensor([[0, 1, 0], [1, 0, 0]]), Nx.tensor([[0, 1, 0], [0, 1, 0]]))
  #Nx.Tensor<
    f32
    0.5
  >
```

Do you want anything other than this?

seanmor5:
@alisinabh Nope, that's perfect :)

PR #268, opened 2022-06-06
Update Nx and EXLA to 0.2.1

alisinabh:


seanmor5:
@alisinabh Thank you for bringing this to my attention! I have fixed on main, which I believe means you can update your previous PR updating the examples :)

Issue #244, opened 2022-05-05
Move model_state inside JIT in training loop


seanmor5:

noozo:
Is axon unusable atm? The credit_card_fraud example doesn't work because of this (at least using EXLA). 

Is there a way to work around it while this issue persists?

josevalim:
Axon has not been fully updated to Nx+EXLA v0.2 yet, which is why you may run into bugs. If you rollback to previous EXLA versions on Git (the ones before EXLA.Backend were introduced), then it should work.

noozo:
Guess i'll wait then. Might as well have my new code with EXLA 0.2 already. Thanks!

alisinabh:
Now I'm getting this error in the evaluator Loop in `recur_predict_fun`. I cannot get any Loop evaluator to work with EXLA.

```
** (RuntimeError) cannot invoke Nx function because it relies on two incompatible tensor implementations: 
Nx.Defn.Expr and EXLA.Backend. You may need to call Nx.backend_transfer/1 (or Nx.backend_copy/1) 
on one or both of them to transfer them to a common implementation
    (nx 0.3.0-dev) lib/nx/shared.ex:447: Nx.Shared.pick_struct/2
    (nx 0.3.0-dev) lib/nx.ex:3238: Nx.element_wise_bin_op/4
    (axon 0.1.0-dev) lib/axon/compiler.ex:379: anonymous fn/15 in Axon.Compiler.recur_predict_fun/3
    (axon 0.1.0-dev) lib/axon/compiler.ex:192: Axon.Compiler.call_predict_cache/6
    (axon 0.1.0-dev) lib/axon/compiler.ex:336: anonymous fn/8 in Axon.Compiler.recur_predict_fun/3
    (elixir 1.13.4) lib/enum.ex:1715: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
    (axon 0.1.0-dev) lib/axon/compiler.ex:334: anonymous fn/15 in Axon.Compiler.recur_predict_fun/3
    (axon 0.1.0-dev) lib/axon/compiler.ex:192: Axon.Compiler.call_predict_cache/6
```

josevalim:
@alisinabh can you provide a way to reproduce the issue? :)

josevalim:
Also please make sure to use latest Axon too!

alisinabh:
@josevalim You can try `examples/vision/cifar10.exs` example. The the evaluation loop is not working because of this error.
[...]

PR #270, opened 2022-06-07
Model State

seanmor5:
I believe this should be the final fix for #244 

Not sure if it's the best approach because we will have large parameters on the backend, and also be storing them as constants in a large program. I think there might be a more efficient way to just pass the params in and still jit the whole eval step

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -384,15 +384,17 @@ defmodule Axon.Loop do
 
     init_fn = fn ->
       %{
+        model_state: model_state,
         y_true: Nx.tensor(0.0),
         y_pred: Nx.tensor(0.0)
       }
     end
 
-    step_fn = fn {inp, tar}, _ ->
+    step_fn = fn {inp, tar}, %{model_state: model_state} ->
       %{
+        model_state: model_state,
         y_true: tar,
-        y_pred: forward_model_fn.(Nx.backend_copy(model_state), inp)
+        y_pred: forward_model_fn.(model_state, inp)
       }
     end
 
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -392,7 +392,7 @@ defmodule Axon.Loop do
     step_fn = fn {inp, tar}, _ ->
       %{
         y_true: tar,
-        y_pred: forward_model_fn.(model_state, inp)
+        y_pred: forward_model_fn.(Nx.backend_copy(model_state), inp)
       }
     end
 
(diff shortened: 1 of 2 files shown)


josevalim:
Yes, I _think_ this is problematic because:

1. It will force each step to JIT again because it has a different closure
2. We will copy the tensor data

I believe the correct fix is to add a new parameter to the training step. We need both parameters and model state (think about it as a map reduce, rather than a reduce). If that's a large change, then you can hide inside the parameter as you get data in and out of the training steps.

josevalim:
How many changes would be required to change `step_fn` (and related functions) to explicitly receive `model_state` as a parameter? :)

seanmor5:
@josevalim I've pushed another fix which might be an option

lib/axon/loop.ex
```diff
@@ -384,15 +384,17 @@ defmodule Axon.Loop do
 
     init_fn = fn ->
       %{
+        model_state: model_state,
```
josevalim, line 387: Is `init_fn` jitted? If so, we have the same issue that we are passing a `EXLA.Backend` tensor to `defn` behind the scenes, which means that it will be converted to binary and inlined into the code. Is this what we want? It feels like we should be raising on said implicit conversions. Should we change Nx to raise?

In any case, I like the direction of moving the concern to `init_fn`. What if we actually pass the model state as argument to `init_fn`? This would build on top of this change and address all of my concerns.

If we do so, I would also change Nx to raise if we try to convert any non binary tensor to an expression (because it almost always means that we are passing something as a closure when we should not).

WDYT?

josevalim:
In other words, I am proposing to go from this:

    model
    |> Axon.Loop.evaluator(model_state)
    |> Axon.Loop.metric(:accuracy, "Accuracy")
    |> Axon.Loop.run(Stream.zip(test_images, test_labels), compiler: EXLA)

to this:

    model
    |> Axon.Loop.evaluator()
    |> Axon.Loop.metric(:accuracy, "Accuracy")
    |> Axon.Loop.run(model_state, Stream.zip(test_images, test_labels), compiler: EXLA)

---

There is also another option which is to not JIT `init_fn` but I am not sure of the remaining consequences of that. You can either not JIT it or force it to not be a function (therefore a regular map).

lib/axon/loop.ex
josevalim: > If we do so, I would also change Nx to raise if we try to convert any non binary tensor to an expression (because it almost always means that we are passing something as a closure when we should not).

I added this check and caught bugs on our EXLA.Backend implementation, so I would say this is the way to go!

PR #274, opened 2022-06-07
Introduce metadata to Axon

seanmor5:
For now the only metadata that is used internally is `op` for custom layers, but this could be useful to external libraries as well

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -90,9 +90,9 @@ defmodule Axon do
   @derive {
     Nx.Container,
     containers: [],
-    keep: [:id, :name, :output_shape, :parent, :parameters, :args, :op, :policy, :hooks, :opts]
+    keep: [:id, :name, :output_shape, :parent, :parameters, :args, :op, :policy, :hooks, :opts, :meta]
   }
-  defstruct [:id, :name, :output_shape, :parent, :parameters, :args, :op, :policy, :hooks, :opts]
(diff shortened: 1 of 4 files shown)


lib/axon.ex
```diff
@@ -132,10 +158,11 @@ defmodule Axon do
 
     {name, opts} = Keyword.pop(opts, :name)
     {shape, opts} = Keyword.pop(opts, :shape)
+    {op_name, opts} = Keyword.pop(opts, :op_name)
```
jonatanklosko, line 161:
```suggestion
    {op_name, opts} = Keyword.pop(opts, :op_name, :custom)
```

lib/axon.ex
```diff
@@ -2783,11 +2892,10 @@ defmodule Axon do
...
-        case {op, opts[:layer_op]} do
-          {op, _} when is_atom(op) -> Atom.to_string(op)
-          {_, nil} -> "custom"
-          {_, layer_op} when is_atom(layer_op) -> Atom.to_string(layer_op)
-          {_, layer_op} when is_binary(layer_op) -> layer_op
+        if op_name do
+          Atom.to_string(op_name)
+        else
+          "custom"
         end
```
jonatanklosko, line 2899: If we have a default, we don't need to worry about checking `nil` consistently :)

jonatanklosko (approved):
Looks great, just the CI is not happy :cat:

josevalim (approved):

PR #279, opened 2022-06-14
Validate options in top-level functions

jonatanklosko:
It's easy to use a wrong option name (`:eps` vs `:epsilon`, `:stride` vs `:strides`), which results in wrong output and is hard to track down. This PR adds `Keyword.validate!` to all functions in the `Axon` module to prevent that.

Also:

* adds missing `:use_bias` in `conv_lstm`
* fixes `:use_bias` handling in `separable_conv2d`, `separable_conv3d`, `lstm` and `gru`
* normalizes the `:axis` option to `concatenate`
* unifies option docs

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -127,9 +127,9 @@ defmodule Axon do
   specifies inference options. All options passed to layer are forwarded
   to inference function except:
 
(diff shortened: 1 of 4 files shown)


seanmor5 (approved):
This is fantastic! Thank you :)

PR #295, opened 2022-06-22
Allow containers as inputs

### LINKED ISSUE #293: Support containers in inputs
Some architectures in other frameworks make use of containers as inputs to simplify APIs. We can and should support a similar construct

seanmor5:
Resolves #293

### DIFF
--- a/lib/axon/compiler.ex
+++ b/lib/axon/compiler.ex
@@ -272,16 +272,7 @@ defmodule Axon.Compiler do
             inputs
 
           %{} = inputs ->
-            cond do
-              Map.has_key?(inputs, name) ->
-                inputs[name]
-
-              is_container_shape(shape) ->
-                inputs
-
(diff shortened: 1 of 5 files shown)


lib/axon/compiler.ex
```diff
@@ -558,4 +565,70 @@ defmodule Axon.Compiler do
...
+  end
+
+  defp is_container_shape(shape) when is_map(shape), do: true
+
+  defp is_container_shape(shape) when is_tuple(shape) do
+    if tuple_size(shape) == 0 do
+      true
+    else
+      not is_dim(elem(shape, 0))
+    end
```
josevalim, line 629:
```suggestion
    tuple_size(shape) == 0 or not is_dim(elem(shape, 0))
```

josevalim (approved):

PR #300, opened 2022-06-24
Add LSTM text generator example and livebook

alisinabh:
This example is based on [this post](https://machinelearningmastery.com/text-generation-lstm-recurrent-neural-networks-python-keras/) by Jason Brownlee.

### DIFF
--- a/examples/generative/text_generator.exs
+++ b/examples/generative/text_generator.exs
@@ -1,0 +1,115 @@
+# Based on https://machinelearningmastery.com/text-generation-lstm-recurrent-neural-networks-python-keras/
+Mix.install([
+  {:axon, github: "elixir-nx/axon"},
+  {:nx, "~> 0.2.1"},
+  {:exla, "~> 0.2.2"},
+  {:req, "~> 0.3.0"}
+])
+
+EXLA.set_as_nx_default([:tpu, :cuda, :rocm, :host])
+
+defmodule TextGenerator do
+  require Axon
+
+  @download_url "https://www.gutenberg.org/files/11/11-0.txt"
+  @sequence_length 100
+  @batch_size 128
+
+  def build_model(characters_count) do
+    Axon.input({nil, @sequence_length, 1}, "input_chars")
+    |> Axon.lstm(256)
+    |> then(fn {_, out} -> out end)
+    |> Axon.nx(fn t -> t[[0..-1//1, -1]] end)
+    |> Axon.dropout(rate: 0.2)
+    |> Axon.dense(characters_count, activation: :softmax)
+  end
+
+  def generate(model, params, init_seq, char_to_idx, idx_to_char, characters_count) do
+    init_seq =
+      init_seq
+      |> String.downcase()
+      |> String.to_charlist()
+      |> Enum.map(&Map.fetch!(char_to_idx, &1))
+
+    Enum.reduce(1..80, init_seq, fn _, seq ->
+      init_seq =
+        seq
+        |> Enum.take(-@sequence_length)
+        |> Nx.tensor()
+        |> Nx.divide(characters_count)
+        |> Nx.reshape({1, @sequence_length, 1})
+
+      char =
+        Axon.predict(model, params, init_seq)
+        |> Nx.argmax()
+        |> Nx.to_number()
+
+      seq ++ [char]
+    end)
+    |> Enum.map(&Map.fetch!(idx_to_char, &1))
+  end
+
+  def transform_text(text, char_to_idx, characters_count) do
+    train_data =
+      text
+      |> Enum.map(&Map.fetch!(char_to_idx, &1))
+      |> Enum.chunk_every(@sequence_length, 1, :discard)
+      # We don't want the last chunk since we don't have a prediction for it.
+      |> Enum.drop(-1)
+      |> Nx.tensor()
+      |> Nx.divide(characters_count)
+      |> Nx.reshape({:auto, @sequence_length, 1})
+      |> Nx.to_batched_list(@batch_size)
+
+    train_labels =
+      text
+      |> Enum.drop(@sequence_length)
+      |> Enum.map(&Map.fetch!(char_to_idx, &1))
+      |> Nx.tensor()
+      |> Nx.reshape({:auto, 1})
+      |> Nx.equal(Nx.iota({characters_count}))
+      |> Nx.to_batched_list(@batch_size)
+
+    {train_data, train_labels}
+  end
+
+  def run do
+    normalized_book_text =
+      Req.get!(@download_url).body
+      |> String.downcase()
+      |> String.replace(~r/[^a-z \.\n]/, "")
+      |> String.to_charlist()
+
+    # Extract all then unique characters we have. Optionally we can sort them.
+    characters = normalized_book_text |> Enum.uniq() |> Enum.sort()
+    characters_count = Enum.count(characters)
+    # Create a mapping for every character
+    char_to_idx = Enum.with_index(characters) |> Enum.into(%{})
+    idx_to_char = Enum.with_index(characters, &{&2, &1}) |> Enum.into(%{})
+
+    model = build_model(characters_count)
+    IO.inspect(model)
+
+    {train_data, train_labels} =
+      transform_text(normalized_book_text, char_to_idx, characters_count)
+
+    IO.puts("Total batches: #{Enum.count(train_data)}")
+
+    params =
+      model
+      |> Axon.Loop.trainer(:categorical_cross_entropy, Axon.Optimizers.adam(0.001))
+      |> Axon.Loop.run(Stream.zip(train_data, train_labels), %{}, epochs: 20, compiler: EXLA)
+
+    init_sequence = """
+    not like to drop the jar for fear
+    of killing somebody underneath so managed to put it into one of the
+    cupboards as she fell past it.
+    """
+
+    generated = generate(model, params, init_sequence, char_to_idx, idx_to_char, characters_count)
+
+    IO.puts("Generated text: #{generated}")
+  end
+end
+
+TextGenerator.run()
--- a/notebooks/text_generator.livemd
+++ b/notebooks/text_generator.livemd
@@ -1,0 +1,580 @@
+# Text Generation with LSTM in Axon
+
+```elixir
+Mix.install([
+  {:axon, github: "elixir-nx/axon"},
+  {:nx, "~> 0.2.1"},
+  {:exla, "~> 0.2.2"},
+  {:req, "~> 0.3.0"}
+])
+
+EXLA.set_as_nx_default([:tpu, :cuda, :rocm, :host])
+```
+
+<!-- livebook:{"output":true} -->
+
+```
+:cuda
+```
+
+## Recurrent Neural Networks as generative models
+
+Recurrent neural networks can also be used as generative models.
+
+This means that in addition to being used for predictive models (making predictions) they can learn the sequences of a problem and then generate entirely new plausible sequences for the problem domain.
+
+Generative models like this are useful not only to study how well a model has learned a problem, but to learn more about the problem domain itself.
+
+In this example, we will discover how to create a generative model for text, character-by-character using LSTM recurrent neural networks in Elixir with Axon.
+
+## Preparation
+
+Using [Project Gutenburg](https://www.gutenberg.org/) we can download a text book which is no longer protected under copywrite so we can experiment with them.
+
+The one that we will use for this experiment is [Alice's Adventures in Wonderland by Lewis Carroll](https://www.gutenberg.org/ebooks/11). You can choose any other text or book that you like for this experiment.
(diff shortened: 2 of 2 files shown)


examples/generative/text_generator.exs
```diff
@@ -0,0 +1,124 @@
...
+      init_seq
+      |> String.downcase()
+      |> String.to_charlist()
+
+    Enum.reduce(1..80, init_seq, fn _, seq ->
+      init_seq =
+        seq
+        |> Enum.reverse()
+        |> Enum.take(@sequence_length)
+        |> Enum.reverse()
```
seanmor5, line 49: I am confused why you need the double reverse?

seanmor5 (approved):
Wow, awesome work! Can you by chance also contribute this as a Livebook with step-by-step explanations and saved outputs? That way we can include it in the docs. It's okay if it's in a separate PR

examples/generative/text_generator.exs
jonatanklosko, line 49:
```suggestion
        |> Enum.take(-@sequence_length)
```

examples/generative/text_generator.exs
```diff
@@ -0,0 +1,124 @@
...
+        |> Enum.map(&Map.fetch!(char_to_int, &1))
+        |> Nx.tensor()
+        |> Nx.divide(Enum.count(characters))
+        |> Nx.reshape({1, @sequence_length, 1})
+
+      char =
+        Axon.predict(model, params, init_seq)
+        |> Nx.argmax()
+        |> Nx.to_number()
+        |> then(&Enum.at(characters, &1))
```
jonatanklosko, line 59: We are converting the number to char, but in the next iteration we convert all characters to numbers. One alternative would be to convert `init_seq` to numbers, work on numbers, and the map back to characters at the end?

examples/generative/text_generator.exs
jonatanklosko: Is this because of the character we add at the end of each iteration? If so, then we can shift it there, like `tl(seq) ++ [char]`.

alisinabh: The reason is I need the last `@sequence_length` tokens for the next input but in correct order. The better approach as mentioned by @jonatanklosko is to use `Enum.take(-@sequence_length)`.

alisinabh: The first suggestion is better because in the second one we will loose some parts of the sentence especially if we generate more than sequence length.

alisinabh: Addressed in 0e6b697

examples/generative/text_generator.exs
alisinabh: Good point. Addressed in 0e6b697

alisinabh:
@seanmor5 Sure. I have actually made the example out of the livebook that I have created :sweat_smile: I've included it in this PR. Please let me know if there are any mistakes or confusion in it.

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,920 @@
...
+filename = "book.txt"
+# change download_url if you like to experiment with other text books.
+download_url = "https://www.gutenberg.org/files/11/11-0.txt"
+
+unless File.exists?(filename) do
+  Req.get!(download_url)
+  |> then(fn %{body: content} -> File.write!(filename, content) end)
+end
+
+book_text = File.read!(filename)
```
jonatanklosko, line 46: Since the payload is not really big, we don't need to cache. Also, with this approach changing the URL wouldn't have effect until the file is deleted :)

```suggestion
# Change the URL if you like to experiment with other text books
download_url = "https://www.gutenberg.org/files/11/11-0.txt"

book_text = Req.get!(download_url).body
```

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,920 @@
...
+```
+
+First of all, we need to normalize the content of the book. We are only interested in the sequence of English characters, periods and new lines. Also currently we don't care about the capitalization and things like apostrophe so we can remove all other unknown characters and downcase everything. We can use regex for that.
```
jonatanklosko, line 55:
```suggestion
First of all, we need to normalize the content of the book. We are only interested in the sequence of English characters, periods and new lines. Also currently we don't care about the capitalization and things like apostrophe so we can remove all other unknown characters and downcase everything. We can use a regular expression for that.
```

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,920 @@
...
+
+We can also convert the string into a list of characters so we can handle them easier. You will understand exactly why a bit further.
+
+```elixir
+normalized_book_text =
+  book_text
+  |> String.downcase()
+  |> String.replace(~r/[^A-Za-z \.\n]/, "")
```
jonatanklosko, line 63: Since we downcase beforehand

```suggestion
  |> String.replace(~r/[^a-z \.\n]/, "")
```

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,920 @@
...
+```
+
+Lowercase English characters in ASCII encoding are presented with numbers between `97 = a` and `122 = z` a space is `32 = [ ]` a new line is `10 = \n` and the period is `46 = .`.
```
jonatanklosko, line 73: More precisely, the charlist includes unicode code points, it's just that in our case it matches ASCII.

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,920 @@
...
+
+```elixir
+sequence_length = 100
+
+train_data =
+  normalized_book_text
+  |> Enum.map(&Map.fetch!(char_to_int, &1))
+  |> Enum.chunk_every(sequence_length, 1, :discard)
+  # We don't want the last chunk since we don't have a prediction for it.
+  |> Enum.slice(0..-2)
```
jonatanklosko, line 185:
```suggestion
  |> Enum.drop(-1)
```

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,920 @@
...
+  [0, 0, 1]
+]
+```
+
+In Nx, you can achieve this encoding by using this snippet
+
+<!-- livebook:{"force_markdown":true} -->
+
+```elixir
+Nx.equal(tensor, Nx.tensor(Enum.to_list(0..last_feature)))
```
jonatanklosko, line 281:
```suggestion
Nx.equal(tensor, Nx.iota({1, last_feature}))
```

Maybe make it an anoonymous function, and show how it works instead of the iex snippet above?

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,920 @@
...
+```elixir
+# As the input, We expect the sequence_length characters. Each one is a number.
+model =
+  Axon.input({nil, sequence_length, 1}, "input_chars")
+  # The Long Short Term Memory layer of our network
+  |> Axon.lstm(256)
+  # Selecting only the output from the LSTM Layer
+  |> then(fn {_, out} -> out end)
+  # Since we only want the last sequence in LSTM we will slice it and select the last one
+  |> Axon.nx(fn x -> Nx.slice_along_axis(x, sequence_length - 1, 1, axis: -2) end)
```
jonatanklosko, line 337: I think if we do this, we no longer need flatten

```suggestion
  |> Axon.nx(fn x -> x[[0..-1//1, -1, 0..-1//1]] end)
```

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,920 @@
...
+train_batches = Nx.to_batched_list(train_data, batch_size)
+result_batches = Nx.to_batched_list(train_results, batch_size)
+
+IO.puts("Total batches: #{Enum.count(train_batches)}")
+
+params =
+  model
+  |> Axon.Loop.trainer(:categorical_cross_entropy, Axon.Optimizers.adam(0.001))
+  |> Axon.Loop.run(Stream.zip(train_batches, result_batches), %{}, epochs: 20, compiler: EXLA)
+```
```
jonatanklosko, line 392: Unless you think showing the parameters makes sense, we can do this:

````suggestion

:ok
```
````

File: notebooks/text_generator.livemd [hunk 11]
```diff
@@ -0,0 +1,920 @@
...
+       ]
+     >}
+  }
+}
+```
+
+## Generating Text
+
+After last step we will have our trained network parameters so we can start generating text with it. We simply need to give the initial sequence as the input to the network and select the most confident output. `Axon.predict/3` will give us the output layer and then using `Nx.argmax/1` we get the most confident neuron index then simply convert that index back to it's ASCII presentation.
```
jonatanklosko, line 527:
```suggestion
After last step we will have our trained network parameters so we can start generating text with it. We simply need to give the initial sequence as the input to the network and select the most confident output. `Axon.predict/3` will give us the output layer and then using `Nx.argmax/1` we get the most confident neuron index then simply convert that index back to it's ASCII representation.
```

(though maybe we should say Unicode code point instead)

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,920 @@
...
+      |> Nx.reshape({1, sequence_length, 1})
+
+    char =
+      Axon.predict(model, params, init_seq)
+      |> Nx.argmax()
+      |> Nx.to_number()
+
+    seq ++ [char]
+  end)
+  |> Enum.map(&Enum.at(characters, &1))
```
jonatanklosko, line 556: We can create `int_to_char` and use that for more efficient mapping.

examples/generative/text_generator.exs
```diff
@@ -0,0 +1,122 @@
...
+    normalized_book_text =
+      book_text()
+      |> String.downcase()
+      |> String.replace(~r/[^A-Za-z \.\n]/, "")
+      |> String.to_charlist()
+
+    # Extract all then unique characters we have. Optionally we can sort them.
+    characters = normalized_book_text |> Enum.uniq() |> Enum.sort()
+    # Create a mapping for every character
+    char_to_int = Enum.with_index(characters) |> Enum.into(%{})
```
jonatanklosko, line 97: It may be confusing that characters are already integers, perhaps if we name it `char_to_idx` it will be more clear?

jonatanklosko (commented):
Awesome, a couple more comments :)

notebooks/text_generator.livemd
alisinabh: This is great. I tested it and it seems the last part is not needed since we want to select everything anyways. 
So `x[[0..-1//1, -1]]` should do the trick.

notebooks/text_generator.livemd
alisinabh: I created an `idx_to_char` mapping for explicitness. But AFAIK it wont help with efficiency as we only have `29` members and I believe BEAM treats maps smaller than `32` elements as lists. But it will make a difference if we expand the codepoints.

notebooks/text_generator.livemd
alisinabh: Addressed in 7a502c4

notebooks/text_generator.livemd
jonatanklosko: > I believe BEAM treats maps smaller than 32 elements as lists

Correct, but the keys are sorted, so lookup via binary search is O(log N) :)

notebooks/text_generator.livemd
jonatanklosko: Yup, all remaining axes are sliced in full!

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,580 @@
...
+
+<!-- livebook:{"output":true} -->
+
+```
+:ok
+```
+
+## Generating Text
+
+After last step we will have our trained network parameters so we can start generating text with it. We simply need to give the initial sequence as the input to the network and select the most confident output. `Axon.predict/3` will give us the output layer and then using `Nx.argmax/1` we get the most confident neuron index then simply convert that index back to it's unicode representation.
```
jonatanklosko, line 369:
```suggestion
After last step we will have our trained network parameters so we can start generating text with it. We simply need to give the initial sequence as the input to the network and select the most confident output. `Axon.predict/3` will give us the output layer and then using `Nx.argmax/1` we get the most confident neuron index then simply convert that index back to it's Unicode representation.
```

jonatanklosko (approved):
:cat:

notebooks/text_generator.livemd
```diff
@@ -0,0 +1,580 @@
...
+```
+
+Lowercase English characters in unicode encoding are presented with numbers between `97 = a` and `122 = z` a space is `32 = [ ]` a new line is `10 = \n` and the period is `46 = .`.
```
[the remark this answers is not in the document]
alisinabh: Addressed in feb8f98

notebooks/text_generator.livemd
alisinabh: Addressed in feb8f98

alisinabh:
Thanks @jonatanklosko 
Squashed and ready!

PR #307, opened 2022-06-26
Fix indentation and inline merging

josevalim:


lib/axon/compiler.ex
```diff
@@ -21,26 +21,19 @@ defmodule Axon.Compiler do
...
-      # behavior is not correct)?
       {time, result} =
         :timer.tc(fn ->
+          params = cache[root_id].(cache, %{})
+
+          # TODO: Maybe merge is not best here, do we want this
+          # operation to fail if structure of params do not match
+          # or should it just silently continue (even though the
+          # behavior is not correct)?
           Map.merge(params, init_params)
```
josevalim, line 32: Unless you have north of millions of keys, this will always be fast, so I squeezed it. :)

PR #308, opened 2022-06-27
Add graph manipulation and utilities API

### LINKED ISSUE #117: Introduce public graph manipulation API
I have been reading a bit into [torch.fx](https://pytorch.org/docs/stable/fx.html) and how that makes it possible to implement custom compiler passes and other cool graph manipulations in PyTorch. I would like to expose a somewhat similar API to Axon to allow for easy manipulation of the Axon graph.

The main Axon API is basically a DSL for creating and composing graphs. These graphs (or Axons as I like to call them) have the following (recursive) typedef:

```elixir
@type axon :: Axon.t() | tuple(axon())
```

The axon  data structure is a tree-like data structure. I am proposing to add tree manipulation functions which look something like:

```elixir
tree_lower(graph :: axon(), any(), (axon(), any() -> {any(), axon()})) :: any() # Unsure of name

# Conveniences around the tree_lower
tree_reduce(graph :: axon(), any(), (axon(), any() -> any()) :: any() 
tree_map(graph :: axon(), (axon -> axon)) :: axon()
```

`tree_lower` takes in an entire graph, an initial accumulator, and a function which takes in the current node and the accumulator and returns the updated accumulator and the next node to traverse. Returning the next node can be useful specifically for writing compiler passes in which certain patterns might be fused or changed in some way. The `lower` comes from the fact that this function can be used as the fundamental function for lowering from Axon to Nx or another representation. The Axon compiler will be rewritten in this way, and so will Axon ONNX.

Both `tree_reduce` and `tree_map` are conveniences around the lowering operation that don't have to concern themselves with traversal (how they're traversed is pretty much set in place). `tree_map` will be used to manipulate node manipulations such as parameter freezing and mixed precision. `tree_reduce` will be used for graph convenience methods like `get_parameters` and `get_inputs`.

I think this will kind of help unify the implementations of AxonOnnx and the Axon Nx Expr compiler. I am considering other abstractions around the Axon compiler, but this is the one I wanted to expose first.

seanmor5:
Resolves #117 

I added a bunch of additional metadata utilities, since there was no public API for that yet and exposing an API makes it more clear how to manipulate parts of the Axon struct. Additionally, it allows us to validate internally to avoid messed up graphs and things going wrong at execution time

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -3063,25 +3063,25 @@ defmodule Axon do
   """
   def freeze(%Axon{} = model, fun \\ & &1) when is_function(fun, 1) do
     parameters =
-      tree_reduce(model, MapSet.new(), fn %Axon{parameters: params}, acc ->
-        Enum.reduce(params, acc, fn param, acc ->
+      reduce_nodes(model, MapSet.new(), fn %Axon{} = axon, acc ->
+        Enum.reduce(get_parameters(axon), acc, fn param, acc ->
           MapSet.put(acc, param)
         end)
       end)
 
     parameters_to_freeze = fun.(Enum.to_list(parameters))
 
-    tree_map(model, fn %Axon{parameters: params} = axon ->
+    map_nodes(model, fn %Axon{} = axon ->
       frozen_params =
-        Enum.map(params, fn %{name: param_name} = v ->
-          if Enum.any?(parameters_to_freeze, fn %{name: name} -> name == param_name end) do
+        Enum.map(get_parameters(axon), fn %{id: param_id} = v ->
+          if Enum.any?(parameters_to_freeze, fn %{id: id} -> param_id == id end) do
             %{v | frozen: true}
           else
             v
           end
         end)
 
-      %{axon | parameters: frozen_params}
+      set_parameters(axon, frozen_params)
     end)
   end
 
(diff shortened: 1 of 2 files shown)


lib/axon.ex
```diff
@@ -3139,93 +3139,327 @@ defmodule Axon do
...
+    %Axon{output_shape: parent_shape} = parent
+    # TODO: compatible? does not handle container shapes
+    unless Axon.Shape.compatible?(parent_shape, shape) do
+      raise ArgumentError,
+            "new parent graph must be compatible with old parent" <>
+              " graph, but got incompatible shapes #{inspect(parent_shape)}" <>
+              " and #{inspect(shape)}"
+    end
 
-  def tree_reduce(%Axon{parent: x} = axon, acc, fun) when is_list(x) do
```
seanmor5, line 3164: I guess I should probably deprecate this instead?

josevalim: Nah, it is fine. :) It is very early, best to move forward.

josevalim:
Looks great! My only suggestion would be to tag those functions as `@doc type: :traversal` perhaps? Or alternatively move them to an `Axon.Graph` module or similar, in case `Axon` is growing too large.

seanmor5:
@josevalim I may choose to reorganize the API so some functions are in different modules before the next release, but for now I am okay with leaving them at the top-level :)

PR #309, opened 2022-06-29
add top-k categorical accuracy

cigrainger:
This is pretty useful when training multi class classification problems where you're happy with the correct answer being in the top k rather than being the argmax.

See [pytorch ignite](https://pytorch.org/ignite/generated/ignite.metrics.TopKCategoricalAccuracy.html).

Alternatively, it could be implemented as an opt on `Metrics.accuracy` like in [torchmetrics](https://torchmetrics.readthedocs.io/en/stable/classification/accuracy.html?highlight=top-k#torchmetrics.Accuracy.params.top_k).

### DIFF
--- a/lib/axon/metrics.ex
+++ b/lib/axon/metrics.ex
@@ -371,6 +371,8 @@ defmodule Axon.Metrics do
   ## Options
 
     * `k` - The k in "top-k". Defaults to 5.
+    * `sparse` - If `y_true` is a sparse tensor. Defaults to `false`.
+
 
   ## Argument Shapes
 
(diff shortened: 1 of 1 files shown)


lib/axon/metrics.ex
```diff
@@ -365,6 +365,44 @@ defmodule Axon.Metrics do
...
+
+      iex> Axon.Metrics.top_k_categorical_accuracy(Nx.tensor([[0, 1, 0], [1, 0, 0]]), Nx.tensor([[0.1, 0.4, 0.7], [0.1, 0.4, 0.7]]), k: 2)
+      #Nx.Tensor<
+        f32
+        0.5
+      >
+
+  """
+  defn top_k_categorical_accuracy(y_true, y_pred, opts \\ []) do
+    opts = keyword!(opts, k: 5)
```
seanmor5, line 396: Maybe we can add a `:sparse` case as well? If `sparse: true` you don't need to compute the the argmax of `y_true` and instead just pass the `y_true` direct

cigrainger: Oh good point! I'll add it.

cigrainger: Okay maybe this is a silly question but how would you go about handling that conditional? I'm struggling to get past `boolean predicate passed to cond` in defn. Would it be reasonable to make the function a normal `def` function that delegates to `defnp` after assigning the index based on that conditional?

polvalente: Perhaps taking a look into Nx.LinAlg.norm or Nx.LinAlg.determinant might provide insight on this conditional implementation :)

cigrainger: Interesting! So `norm` is defn that delegates to defp and `determinant` is def that delegates to defnp. I think I'm more confused than where I started now lol.

josevalim: The issue is that you cannot execute regular code inside a `defn`, so you need to do something like this:

```
transform(opts[:flag], fn flag ->
  if flag do
    call_another_defn_for_flag(...)
  else
    call_a_defn_without_flag(...)
  end
end)
```

josevalim: We definitely need to improve this though...

cigrainger: Alright, how's that now?

cigrainger:
Nah this is wrong -- I was getting mixed up with argsort. Argsort gives the indices needed to sort the array in order. I need to replace each value with its index in a sorted array.

cigrainger:
Okay @seanmor5 sorry for the delay in updating this. How's this look?

seanmor5 (approved):
This looks great to me! Thank you :)

PR #310, opened 2022-07-01
Lazy model compilation

seanmor5:
WIP: Adds a lazy compilation which requires a template shape to build the model. There are lots of changes that still need to be made:

1) Blow up most shape functions aside from param shapes
2) Probably completely remove `output_shape` from the layer altogether
3) Unify layer handling (e.g. we can just pass the underlying layer function as `op` now)
4) Move table inspect to a display module and re-write traditional inspect


lib/axon.ex
```diff
@@ -3495,8 +3487,36 @@ defmodule Axon do
   or backend.
   """
   @doc type: :compilation
-  def compile(model, opts \\ []) when is_list(opts) do
-    {Axon.Compiler.compile_init(model, opts), Axon.Compiler.compile_predict(model, opts)}
+  def build(model, opts \\ []) when is_list(opts) do
+    {init_fn, predict_fn} = Axon.Compiler.build(model, opts)
+    opts = [on_conflict: :reuse] ++ opts
+    # TODO: Wrap in jit_apply
```
josevalim, line 3493: Yes, the `on_conflict: :reuse` assumes you are using Nx v0.3, which means Nx.Defn.jit/3 should be printing deprecation warnings. :)

lib/axon.ex
```diff
@@ -3495,8 +3487,36 @@ defmodule Axon do
...
+  inputs which match the expected shapes and types of inputs at
+  execution time.
+
+  This function makes use of the built-in `Nx.Defn.compile/3`. Note
+  that passing inputs which differ in shape or type from the templates
+  provided to this function will result in potentially expensive
+  recompilation.
+  """
+  @doc type: :compilation
+  def compile(model, template, init_params \\ %{}, opts \\ []) when is_list(opts) do
```
josevalim, line 3512: I am afraid this won't play well with optional model parameters. When you have optional model parameters, then you need to train the function for those precise model parameters and the default model state won't have them, correct?

I don't even know if `init_params` itself can have optional parameters... but we may consider the fact we only want to compile `predict`?

```elixir
def compile(model, template, opts \\ [])

def compile(model, template, opts) when is_list(opts) do
  compile(model, template, nil, opts)
end

def compile(model, template, model_params) when is_map(model_params) do
  compile(model, template, model_params, [])
end

def compile(model, template, model_params, opts) do
  {init_fn, predict_fn} = build(model, opts)
  Nx.Defn.compile(predict_fn, [model_params || init_fn.(template, init_params), template])
end
```

josevalim: The code above obviously won't work, because we don't have `init_params` on the fallback case. So maybe we can't allow compiling without the model parameters at all.

lib/axon.ex
```diff
@@ -3495,8 +3487,36 @@ defmodule Axon do
   or backend.
   """
   @doc type: :compilation
-  def compile(model, opts \\ []) when is_list(opts) do
-    {Axon.Compiler.compile_init(model, opts), Axon.Compiler.compile_predict(model, opts)}
+  def build(model, opts \\ []) when is_list(opts) do
+    {init_fn, predict_fn} = Axon.Compiler.build(model, opts)
+    opts = [on_conflict: :reuse] ++ opts
+    # TODO: Wrap in jit_apply
+    {&Nx.Defn.jit(init_fn, [&1, &2], opts), &Nx.Defn.jit(predict_fn, [&1, &2], opts)}
```
josevalim, line 3494: And you either want to move the Nx.Defn.jit call to the Compiler... or keep it here but then `Keyword.pop(opts, :mode)` here.

lib/axon.ex
josevalim: I am 100% wrong. Just ignore me. :)

PR #317, opened 2022-07-07
Copy transfers early on

josevalim:


josevalim:
I am actually not sure if this is a good idea. :D

josevalim:
Fixed in the upcoming v0.2 branch!

PR #321, opened 2022-07-17
Freeze names in serialization

### LINKED ISSUE #316: Axon deserialize issue: "..is invalid, likely because it points to an old version of the code"
I've been running into some de/serialization issues with the model. It may be that I have loaded or saved the model wrong (just used File.write!/read! defaults.

```
params = 
    model
    |> Axon.Loop.trainer(loss_fn, optimizer)
    |> Axon.Loop.run(data, %{}, epochs: 5, iterations: 1)
    
serialized_model = Axon.serialize(model, params)
File.write!("./model_b", serialized_model)

serialized = File.read!("./model_b")
{saved_model, saved_params} = Axon.deserialize(serialized)
```
After inspecting deserialize/2 in Axon, it seems its it's throwing this error from the map_to_axon call.
```

%Inspect.Error{
  message: "got BadFunctionError with message \"function #Function<58.92691295/2 in Axon> is invalid, likely because it points to an old version of the code\" while inspecting %{__struct__: Axon, args: [:layer], hooks: [], id: 20, name: #Function<57.92691295/2 in Axon>, op: :softmax, op_name: :softmax, opts: [], output_shape: {nil, 1, 9}, parameters: [], parent: [%{__struct__: Axon, args: [:layer, :parameter, :parameter], hooks: [], id: 19, name: #Function<57.92691295/2 in Axon>, op: :dense, op_name: :dense, opts: [], output_shape: {nil, 1, 9}, parameters: [%Axon.Parameter{frozen: false, id: 17, initializer: :glorot_uniform, name: \"kernel\", shape: {100, 9}}, %Axon.Parameter{frozen: false, id: 18, initializer: :zeros, name: \"bias\", shape: {9}}], parent: [%{__struct__: Axon, args: [:layer], hooks: [], id: 16, name: #Function<57.92691295/2 in Axon>, op: :dropout, op_name: :dropout, opts: [rate: 0.2], output_shape: {nil, 1, 100}, parameters: [], parent: [%{__struct__: Axon, args: [:layer], hooks: [], id: 15, name: #Function<38.92691295/2 in Axon>, op: #Function<41.92691295/2 in Axon>, op_name: :elem, opts: [], output_shape: {nil, 1, 100}, parameters: [], parent: [%{__struct__: Axon, args: [...], ...}], policy: p=f32 c=f32 o=f32}], policy: p=f32 c=f32 o=f32}], policy: p=f32 c=f32 o=f32}], policy: p=f32 c=f32 o=f32}"
}
```
This is only happening if you save the Axon.serialize using File.write! in another process, once you run a new process and load the relevant file output you get the below. 

If you save the serialized state and load it straight away (from the same process, or mix run -e) there isn't any issue.

seanmor5:
Resolves #316 

We freeze names and then rebuild the function to avoid trying to serialize functions. Custom layers will still present an issue, but I think fixing that might require a different solution

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -3393,21 +3393,25 @@ defmodule Axon do
   """
   @doc type: :model
   def serialize(%Axon{} = model, params, opts \\ []) do
-    model_meta = axon_to_map(model)
+    {model_meta, _op_counts} = axon_to_map(model, %{})
     params = Nx.serialize(params, opts)
     :erlang.term_to_binary({@file_version, model_meta, params}, opts)
   end
 
-  defp axon_to_map(%Axon{op: :container, parent: [parents]} = model) do
-    parents = deep_new(parents, &axon_to_map/1)
+  defp axon_to_map(%Axon{op: :container, name: name_fn, parent: [parents]} = model, op_counts) do
+    {parents, op_counts} = deep_map_reduce(parents, op_counts, &axon_to_map/2)
     axon_map = Map.from_struct(model) |> Map.put(:axon, :axon)
-    %{axon_map | parent: List.wrap(parents)}
+    name = name_fn.(:container, op_counts)
+    op_counts = Map.update(op_counts, :container, 1, fn x -> x + 1 end)
+    {%{axon_map | parent: List.wrap(parents), name: name}, op_counts}
   end
 
-  defp axon_to_map(%Axon{parent: parents} = model) do
-    parents = Enum.map(parents, &axon_to_map/1)
+  defp axon_to_map(%Axon{op_name: op, parent: parents, name: name_fn} = model, op_counts) do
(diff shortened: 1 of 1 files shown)


josevalim:
For custom layers we need to either support MFArgs tuples or force people to pass functions in the &Mod.fun/arity format.

josevalim (approved):

seanmor5:
@josevalim Do you think it's enough to document that in order to serialize custom layers you need to use the fully qualified MFA function? We could probably also raise in unsupported cases?

josevalim:
I think we should warn if you try to serialize something with "improper layers". I am, however, concerned that `&Mod.fun/3` won't be enough and we will also have to support MFArgs tuples (`{mod, fun, additional_args}`).

For example, if we want to serialize the models in Bumblebee, which we likely want, they will have to be written in said format... so I am even thinking if we should enforce MFArgs altogether.

PR #322, opened 2022-07-18
Redesign optional inputs

jonatanklosko:
We decided to drop input defaults and provide an API around optional inputs. This approach provides more flexibility in computing the defaults, as they can now depend on any other nodes. It also allows for more conditional constructs.

This PR adds the minimal building blocks, that is `Axon.input(..., default: true)`, `%Axon.None{}` propagation and `Axon.optional` for handling `%Axon.None{}` explicitly in layers. Going forward, we will add more layers for dealing with optional subgraphs.


lib/axon.ex
```diff
@@ -363,6 +356,60 @@ defmodule Axon do
...
+  In case the mask is not provided, the input node will resolve to
+  `%Axon.None{}` and so will all the layers that depend on it. By
+  using `optional/2` a layer may opt-in to receive `%Axon.None{}`
+  and it needs to handle it accordingly. To fix our example, we
+  could define a custom layer to apply the mask only when present
+
+      def apply_optional_mask(%Axon{} = x, %Axon{} = mask) do
+        Axon.layer(
+          fn x, mask, _opts ->
+            case mask do
```
seanmor5, line 384: To be clear, we have `case` in `defn` now right? Just making sure this isn't only a valid implementation for anonymous functions as they have their own issues with serialization

seanmor5 (approved):
Beautiful! A few minor comments, otherwise everything looks great :)

lib/axon.ex
jonatanklosko: Currently to use `case` we would need to match on `x.__struct__`, or use a transform. @josevalim the plan is to add basic struct support to make this work, right?

In this context I wonder if we should derive `Nx.Container`, it's not strictly necessary because we don't expect `Axon.None` outside our compilation. On he other hand, if write a separate `defn` that considers `Axon.None` as an argument, it's kinda weird if it doesn't work on a direct call. Opinions?

seanmor5: I am a bit worried about cases where the implementation doesn't work as expected outside of `Axon.predict`, just because I prefer less magic and would rather an implementation like a custom layer work as-is regardless of context. Additionally, I think we'll definitely want `case` in `defn` if we are going to end up matching on `%Axon.None`

jonatanklosko: So deriving `Nx.Container` and extending `case` to support structs should address both concerns?

seanmor5: Yes, I believe so. I'm okay if we just make note of extending case for now and move forward with a merge so we can get this integrated in other places :)

jonatanklosko: Sounds good, I opened https://github.com/elixir-nx/nx/issues/841 :)

PR #325, opened 2022-07-21
Start weight sharing API

### LINKED ISSUE #169: Add weight sharing
There is currently no explicit way to use the same weights/layer for separate inputs. For example, in Keras I could do:

```python
dense = tf.keras.layers.Dense(32)

d_x1 = dense(x1)
d_x2 = dense(x2)
```

And it would use the same layer for both operations. I believe this could potentially work for now:

```elixir
x1 = Axon.input({nil, 32})
x2 = Axon.input({nil, 32})

d_x1 = Axon.dense(x1, 64, name: "dense")
d_x2 = Axon.dense(x2, 64, name: "dense")
```

And the parameters would be shared by name. Perhaps there is a way to make this a bit more explicit. This somewhat relates to some of the issues that appeared earlier with GANs and the proposed `Axon.function`

seanmor5:
Resolves #169

Still a few API considerations/questions:

1. Should we offer an explicit `:parameters` option on layer creation?
2. How should we handle partial parameter sharing, e.g. if I just want to share the kernel of an embedding layer with a dense layer how do I express that?
3. What kind of validations do we need in place to ensure we don't pass invalid parameters to share between layers?
4. How do we share parameters between layers where the sharing is valid, but where the param shape function is incompatible? An example is a sharing between a dense layer and a bilinear layer. The dense layer has a parameter shape function with arity-1 for 1 input and the bilinear layer has a shape function with arity-2 for 2 inputs.
5. What happens if we "cut-off" shared portions of the graph? I don't think the compiler will have the cache built correctly if we do in that case?

These are just the few I can think of right now. I'm sure there's more things to figure out from there. But the good news is the compiler works and the change is really simple :)

### DIFF
--- a/lib/axon/compiler.ex
+++ b/lib/axon/compiler.ex
@@ -310,8 +310,6 @@ defmodule Axon.Compiler do
       namespace_params = params[name]
 
       # TODO: How should hooks be handled here?
-      # TODO: I think we can actually handle parameter freezing and access
-      # better here by only forwarding params[namespace] to the child function
       {out, {state, result_cache}} =
         call_predict_cache(parent_id, namespace_params, inputs, state, cache, result_cache)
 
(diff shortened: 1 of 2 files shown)


josevalim:
> Should we offer an explicit :parameters option on layer creation?

IMO set_parameters is enough if it is not very common and we already have a get_parameters API.

> How should we handle partial parameter sharing, e.g. if I just want to share the kernel of an embedding layer with a dense layer how do I express that?

Can we allow a list of keys on `get_parameters(layer, keys \\ :all)`? Then `set_parameters` can see which keys are missing.

> What happens if we "cut-off" shared portions of the graph? I don't think the compiler will have the cache built correctly if we do in that case?

I think we are able to see the original part is no longer and therefore we need to look at the parameter in the "usual" position.

seanmor5:
Too much has changed, and this needs to be revisited with considerations from the other library

PR #326, opened 2022-07-23
Text Classification LSTM Livebook & Example Integration

fantypants:
Hello, 

This is the initial implementation of the LSTM NER Text Classification. I've worked on the live-book & example for the model.


examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+# Test Data: https://github.com/bhuvanakundumani/NER_tensorflow2.2.0/blob/master/data/test.txt
+
+# Train Data: https://github.com/bhuvanakundumani/NER_tensorflow2.2.0/blob/master/data/train.txt
+
+# Validation Data: https://github.com/bhuvanakundumani/NER_tensorflow2.2.0/blob/master/data/valid.txt
+
+Mix.install([
+  {:axon, github: "elixir-nx/axon"},
+  {:nx, "~> 0.2.1"},
+  {:exla, "~> 0.2.2"}
```
seanmor5, line 15: If using Axon GH you should also use these git deps:

```suggestion
  {:exla, "~> 0.3.0-dev", github: "elixir-nx/nx", sparse: "exla"},
  {:nx, "~> 0.3.0-dev", github: "elixir-nx/nx", sparse: "nx", override: true},
```

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+  @batch_size 128
+  @lstm 128
+  @embed_dimension 256
+  @def_split " "
+  @start "-DOCSTART- -X- -X- O\n"
+  @ending "-------------------------------------------------------------"
+  @incorrect [".", "?", "%", ",", ":", ";", "-docstart-"]
+
+  # seqlen
+  # word_count
```
seanmor5, line 35: Guessing just forgot to remove?

```suggestion
```

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+
+  def train(model, data) do
+    model
+    |> Axon.Loop.trainer(&kl_divergence/2, Axon.Optimizers.adam(0.001))
```
seanmor5, line 39: You should be able to just use the atom here:

```suggestion
    |> Axon.Loop.trainer(:kl_divergence, Axon.Optimizers.adam(0.001))
```

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+    |> Axon.Loop.metric(:accuracy, "Accuracy")
+    |> Axon.Loop.run(data, %{}, epochs: 3, iterations: 5)
```
seanmor5, line 41: Only 5 iterations?

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+
+  end
+
+  defp kl_divergence(y_true, y_pred) do
+    Axon.Losses.kl_divergence(y_true, y_pred, reduction: :mean)
+  end
```
seanmor5, line 47: You can remove once you replace with an atom `:kl_divergence` above
```suggestion
```

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+      Axon.input({nil, @sequence_length, 1}, "inputs")
+       |> Axon.embedding(word_count, @sequence_length)
+       |> Axon.nx(fn t ->
+         t[[0..-1//1, -1]] # Dropping out the sized 1 column created from the embed
+       end)
+       |> Axon.spatial_dropout(rate: 0.1)
+       |> Axon.lstm(@lstm)
+       |> then(fn {{new_cell, new_hidden}, out} ->
+         out
+       end)
```
seanmor5, line 59:
```suggestion
    |> elem(1)
```

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+    |> Axon.Loop.run(data, %{}, epochs: 3, iterations: 5)
+
+  end
+
+  defp kl_divergence(y_true, y_pred) do
+    Axon.Losses.kl_divergence(y_true, y_pred, reduction: :mean)
+  end
+
+  def build_model(word_count, label_count) do
+      Axon.input({nil, @sequence_length, 1}, "inputs")
```
seanmor5, line 50: New Axon doesn't require input shapes :)

You can still give it a hint though

```suggestion
      Axon.input("inputs", shape: {nil, @sequence_length, 1})
```

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+       |> Axon.embedding(word_count, @sequence_length)
+       |> Axon.nx(fn t ->
```
seanmor5, line 52: The reason you have sized 1 is because your input shape is `{nil, seq_len, 1}`. My suggestion is to squeeze your inputs such that the new input shape is `{nil, seq_len}` instead, and then you can remove this

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+         t[[0..-1//1, -1]] # Dropping out the sized 1 column created from the embed
+       end)
+       |> Axon.spatial_dropout(rate: 0.1)
```
seanmor5, line 55: I'm not sure this is a good idea---a better approach would be to use recurrent dropout, but Axon does not support that yet :)

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+    train_data =
+      word_labels
+      |> Enum.map(fn {word, label} ->
+        encode(word_to_idx, word)
+      end)
+      |> Enum.chunk_every(@sequence_length, 1, :discard)
+      |> Enum.drop(-1)
+      |> Nx.tensor
+      |> Nx.divide(wcount)
+      |> Nx.reshape({:auto, @sequence_length, 1})
```
seanmor5, line 92: Relates to shape comment above :)

```suggestion
      |> Nx.reshape({:auto, @sequence_length})
```

seanmor5 (commented):
Nice work! I've left some comments. Also, I recommend running the formatter on the script, I think there are some minor formatting things :)

examples/classifcation/text_classification.exs
fantypants: I like this feedback a lot, using spatial_dropout results in some strange things happening, I've been doing testing on this. So you propose recurrent dropout? I'll have to look into what it is and see what we can do with it. The spatial dropout I believe seems to cause issues after X epochs, I assume it's due to it dropping same channel across the tensors, which could be the feature map.

examples/classifcation/text_classification.exs
fantypants: This is why we learn from you haha

examples/classifcation/text_classification.exs
fantypants: perfect!

examples/classifcation/text_classification.exs
fantypants: Oversight from early testing, I've been running at 100,300,500,600,800 iterations at current, way back at the start i wanted to see what was actually happening with the model loading & saving etc, was just easy for testing. This is something I need to also cover a bit more in the live book as it's an important key factor to understand I believe (I'm completely self taught other than what I've learnt from yourself, recent mentoring in the space). I'll need to do some more research into how elaborate on the hyper-parameters as a whole

examples/classifcation/text_classification.exs
fantypants: So this was a weird bit for me, I followed alot of tutorials at the start for the basic concept, and the sizing were always not well explained. I'll do as suggested above, as it makes more sense, I believe the problem at the start I was having was due to the embedding layer and me not understanding I needed to squeeze it out earlier.

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+  end
+
+  defp encode(dictionary, word) do
+    with {:ok, encoded} <- Map.fetch(dictionary, word) do
+      encoded
+    else
+      _->
+      0
+    end
+  end
```
jonatanklosko, line 71:
```suggestion
  defp encode(dictionary, word), do: dictionary[word] || 0
```

Sidenote: instead of with like this, you can use `case` :)

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+
+  defp decode(dictionary, word) do
+    with {:ok, decoded} <- Map.fetch(dictionary, word) do
+      decoded
+    else
+      _-> "Unknown"
+    end
+  end
```
jonatanklosko, line 79:
```suggestion
  defp decode(dictionary, word), do: dictionary[word] || "Unknown"
```

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+    |> Stream.map(fn line ->
+      [token | entities] = String.split(line, @def_split)
+      if(
+      !is_nil(token) &&
+      token !== "" &&
+      String.length(token) !== 1 &&
+      line !== "\n" &&
+      token !== @start &&
+      token !== @ending &&
+      token not in @incorrect)
```
jonatanklosko, line 181:
```suggestion
      if token not in ["", @start, @ending] and token not in @incorrect and String.length(token) !== 1
```

I don't think `token` can ever be `nil` :)

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+      do
+
+        [ r | rest ] = entities |> List.last |> String.split("\n")
```
jonatanklosko, line 184: We are processing lines from the stream, right? Why do we need to split on newline here?

examples/classifcation/text_classification.exs
```diff
@@ -0,0 +1,273 @@
...
+
+        {String.downcase(token), r}
+      else
+        nil
+      end
+    end)
+    |> Stream.reject(fn x -> is_nil(x) end)
```
jonatanklosko, line 191:
```suggestion
    |> Stream.reject(&is_nil/1)
```

examples/classifcation/text_classification.exs
fantypants: Just validated this & also changed some things based on suggestion. The reason for it there was the weird early handling of the row:
`"WORD LABEL LABEL LABEL\n"` when I adjusted the preprocessing as you suggested below it was much more obvious.

examples/classifcation/text_classification.exs
fantypants: ^Above comment relates to this, all fixed and redone =)

fantypants:
> 

@seanmor5 I've finished the changes and pushed to my main branch, how do I update the PR with the additions?

jdewar:
filename has one too many s's, and directory has one too few i's

`examples/classifcation/text_classsification.exs` -> `examples/classification/text_classification.exs`

examples/classifcation/text_classsification.exs
```diff
@@ -0,0 +1,250 @@
...
+  def train(model, data) do
+    model
+    |> Axon.Loop.trainer(:kl_divergence, Axon.Optimizers.adam(0.001)) # Can use :categorical_cross_entropy
+    |> Axon.Loop.metric(:accuracy, "Accuracy")
+    |> Axon.Loop.run(data, %{}, epochs: 20, iterations: 300)
+  end
+
+  defp kl_divergence(y_true, y_pred) do
+    Axon.Losses.kl_divergence(y_true, y_pred, reduction: :mean)
+  end
```
seanmor5, line 45:
```suggestion
```

I don't think you need this anymore, right?

examples/classifcation/text_classsification.exs
```diff
@@ -0,0 +1,250 @@
...
+    train_labels =
+      word_labels
+      |> Enum.map(fn {word, label} ->
+          encode(label_to_idx, label)
+      end)
+      |> Enum.chunk_every(@sequence_length, 1, :discard)
+      |> Enum.drop(-1)
+      |> Nx.tensor
+      |> Nx.reshape({:auto, @sequence_length, 1})
+      |> Nx.to_batched_list(@batch_size)
```
seanmor5, line 83:
```suggestion
      |> Nx.to_batched(@batch_size)
```

examples/classifcation/text_classsification.exs
```diff
@@ -0,0 +1,250 @@
...
+      word_labels
+      |> Enum.map(fn {word, label} ->
+        encode(word_to_idx, word)
+      end)
+      |> Enum.chunk_every(@sequence_length, 1, :discard)
+      |> Enum.drop(-1)
+      |> Nx.tensor
+      |> Nx.divide(wcount)
+      |> Nx.reshape({:auto, @sequence_length})
+      |> Nx.to_batched_list(@batch_size)
```
seanmor5, line 71:
```suggestion
      |> Nx.to_batched(@batch_size)
```

seanmor5:
@fantypants Hey! Just checking up on the status of this PR? Do you need help or another review?

fantypants:
> @fantypants Hey! Just checking up on the status of this PR? Do you need help or another review?

Hey @seanmor5, Apologies for the delay, I do believe it would be good for a review soon; however, I'll take a look over it tomorrow and fix another highlighted issue with the filename. I also have to go through the changes of late to make sure it's all still compatible. I'll give you a shout on here when it's ready (1-2 Days).

PR #329, opened 2022-07-28
Update Loop.evaluator doc

### LINKED ISSUE #328: Out of date example on Loop.evaluator docs
The documentation for the `Loop.evaluator` function shows an example where two args are required. In the newest version of the lib, the function only accepts one argument instead of two, so the `state` that is shown in the example [here](https://github.com/elixir-nx/axon/blob/878daa579e52ec813f174e7c504cf17999a57421/lib/axon/loop.ex#L607) should be removed.

The example right now is

```elixir
model
|> Axon.Loop.evaluator(trained_state)
|> Axon.Loop.metric("Accuracy", :accuracy)
```

Also there seem to be still more references to `evaluator/2` like [here](https://github.com/elixir-nx/axon/blob/878daa579e52ec813f174e7c504cf17999a57421/lib/axon/loop.ex#L188).

It is just something small in the docs, but reading it on Hex threw me off a bit when trying to use it.

If the only thing that needs to change is to actually remove the extra arg in the example and change the arity in the moduledoc I can open a PR, but if this actually prompts a bigger docs change I don't think I know enough of how it works to do it.

Ian-GL:
Minor changes to update `Loop.evaluator` example in the docs so it is consistent with the actual function.

Fixes #328

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -185,8 +185,7 @@ defmodule Axon.Loop do
       * `Axon.Loop.trainer/3` - Creates a supervised training loop from model, loss, and
       optimizer.
 
-      * `Axon.Loop.evaluator/2` - Creates a supervised evaluator loop from model and model
-      state.
+      * `Axon.Loop.evaluator/1` - Creates a supervised evaluator loop from model.
 
   ## Running loops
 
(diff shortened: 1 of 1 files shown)


seanmor5 (approved):
Thank you :)

PR #341, opened 2022-08-15
Start scan combinator

seanmor5:
WIP: The goal is to replace `*_unroll` entirely with this more general `scan` and to also use it to add both `Axon.scan` and `Axon.bidirectional` combinators

### DIFF
--- a/lib/axon/layers.ex
+++ b/lib/axon/layers.ex
@@ -2453,32 +2453,17 @@ defmodule Axon.Layers do
   may be more efficient for long sequences.
   """
   defn dynamic_unroll(cell_fn, input_sequence, carry, input_kernel, recurrent_kernel, bias) do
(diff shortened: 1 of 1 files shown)


lib/axon/layers.ex
```diff
@@ -2613,4 +2598,65 @@ defmodule Axon.Layers do
 
     Nx.slice_along_axis(input, offset, size, axis: opts[:axis])
   end
+
+  @doc false
+  defn scan(input, acc, fun, env, opts \\ []) do
+    opts = keyword!(opts, reverse: false, unroll: 1, axis: 1)
+
+    initial_shape =
+      transform({fun, input, acc, env}, fn
```
josevalim, line 2607: transform will be deprecated, so you want to avoid it altogether!

josevalim (approved):

seanmor5:
This will be revisited for the next release, closing for now as it is stale!

PR #346, opened 2022-09-06
Warn on non-MFA functions

### LINKED ISSUE #323: Custom layer serialization implementation
Currently custom layers will fail when they are imported if the implementation is not an MFA function (`&Module.function/arity`). We should consider raising in these cases, or support some other way of serializing custom layers. Another option is to trace the custom layer and serialize the resulting Nx expression and then rebuild the function from the expression

seanmor5:
Resolves #323 

For now, just warns. Will raise on next release

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -3428,6 +3428,7 @@ defmodule Axon do
 
   """
   @doc type: :model
+<<<<<<< HEAD
   def serialize(%Axon{output: id, nodes: nodes}, params, opts \\ []) do
     Logger.warning(
       "Attempting to serialize an Axon model. Serialiation is discouraged" <>
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -216,6 +216,8 @@ defmodule Axon do
   alias __MODULE__, as: Axon
   alias Axon.Parameter
 
+  require Logger
+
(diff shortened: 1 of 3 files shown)


seanmor5:
@josevalim I think that's okay because you still wouldn't be able to deserialize the closure correctly in another program?

In cases where you want more flexibility, you can just serialize the parameters and ensure you use the same code to generate the model programmatically

josevalim:
Yes, the closure is wrong either way. My concern is that we cannot reproduce the closure capabilities now, with the chosen API, unless we support an additional optional argument, called `extra_args`, which we append to the existing args. So, for example, we could do:

    layer(x, &SomeFun.bar/4, [y])

seanmor5:
There is a workaround for this which I just pushed, but as you can see it is failing because there are some inconsistencies in generating parameter shapes. Basically if you want a closure you can just use Axon.constant layers as input :)

The issue now is that parameter shapes are functions which depend on the known input shapes at initialization time, which creates kind of weird scenarios where if you swap out a parameter for a constant than it breaks things. I think I need to update that to work regardless of whether a layer input is an Axon input or Parameter input

josevalim:
Perfect then, ship it!

PR #350, opened 2022-09-09
Use RNG Keys

### LINKED ISSUE #31: Support RNG seeds
Necessary for reproducible results.

Requires https://github.com/elixir-nx/nx/issues/333

seanmor5:
Resolves #31 

Blocked on https://github.com/elixir-nx/nx/pull/846

When that PR is ready then we can run with this. Currently you build the model with the seed, so `init_fn` will always return the same parameters.


lib/axon/compiler.ex
```diff
@@ -10,11 +10,13 @@ defmodule Axon.Compiler do
   @doc false
   def build(%Axon{} = graph, opts) do
     debug? = Keyword.get(opts, :debug, false)
-    {mode, _jit_opts} = Keyword.pop(opts, :mode, :inference)
+    {mode, opts} = Keyword.pop(opts, :mode, :inference)
+    {key, _jit_opts} = Keyword.pop(opts, :key, Nx.Random.key(:erlang.system_time()))
```
josevalim, line 14:
```suggestion
    mode = Keyword.get(opts, :mode, :inference)
    key = Keyword.get_lazy(opts, :key, fn -> Nx.Random.key(:erlang.system_time()) end)
```

lib/axon/compiler.ex
```diff
@@ -626,25 +636,25 @@ defmodule Axon.Compiler do
...
+  defp apply_initializer(initializer, shape, type, key) when is_atom(initializer) do
     fun = apply(Axon.Initializers, initializer, [])
-    fun.(shape, type)
+    fun.(shape, type, key)
   end
 
-  defp apply_initializer(initializer, shape, type) when is_function(initializer, 2) do
-    initializer.(shape, type)
+  defp apply_initializer(initializer, shape, type, key) when is_function(initializer, 3) do
+    initializer.(shape, type, key)
```
josevalim, line 657: Shouldn't we have to eventually split the keys? Otherwise all layers will generate the same random numbers. Perhaps support both arity 2 and arity 3 initializers. The one with arity 3 expect a key, which you split before hand?

seanmor5:
@polvalente Any ideas on the Torchx failures?

josevalim:
> @polvalente Any ideas on the Torchx failures?

Fixed!

PR #359, opened 2022-09-15
Start to add some basic guides to the documentation

### LINKED ISSUE #162: Add guides to documentation
Here is an (incomplete) list of guides that should be added to the documentation:

- [x] Basic Axon models
- [x] Complex Models (combinators, weight sharing, etc.)
- [ ] Working with Named Tensors (requires #57)
- [ ] Training and Evaluating Axon models
- [x] Creating Custom Layers
- [ ] Custom Training Loops
- [ ] Creating Custom Event Handlers and Metrics
- [ ] Mixed Precision Training
- [ ] Axon Graph Manipulation (requires #117)
- [ ] Pruning trained models (requires #120)
- [ ] Model Serialization and Saving (requires #26)
- [x] Hooks (requires #77)
- [ ] Axon Optimization Deep Dive
- [ ] Transfer Learning and Fine Tuning

Note these differ from examples of specific applications of Axon as they are intended to be example based deep dives of certain aspects of the API.

seanmor5:
WIP

Resolves #162


guides/model_creation/ComplexModels.livemd
```diff
@@ -0,0 +1,84 @@
+# Complex Models
+
+```elixir
+Mix.install([
+  {:axon, path: "."},
```
jonatanklosko, line 5: We should point to hex before the release, so that the notebook can be run on it's own. For now `{:axon, github: "elixir-nx/axon"}` is probably best :)

mix.exs
```diff
@@ -86,10 +86,16 @@ defmodule Axon.MixProject do
       extras: [
         "notebooks/mnist.livemd",
         "notebooks/fashionmnist_autoencoder.livemd",
-        "notebooks/multi_input_example.livemd"
+        "notebooks/multi_input_example.livemd",
+        "guides/model_creation/YourFirstAxonModel.livemd",
```
jonatanklosko, line 90: Would be better to use snake case for consistency.

guides/model_creation/ComplexModels.livemd
```diff
@@ -0,0 +1,84 @@
...
+```elixir
+input = Axon.input("data")
+
+x1 = input |> Axon.dense(32)
+x2 = input |> Axon.dense(64) |> Axon.relu() |> Axon.dense(32)
+
+out = Axon.add(x1, x2)
+```
+
+In the snippet above, your model branches `input` into `x1` and `x2`. Each branch performs a different set of transformations; however, at the end the branches are merged with an `Axon.add/2`. You might sometimes see layers like `Axon.add/2` called *combinators*. Really they're just layers that operate on multiple Axon models at once---typically to merge some branches together.
```
jonatanklosko, line 24:
```suggestion
In the snippet above, your model branches `input` into `x1` and `x2`. Each branch performs a different set of transformations; however, at the end the branches are merged with an `Axon.add/2`. You might sometimes see layers like `Axon.add/2` called *combinators*. Really they're just layers that operate on multiple Axon models at once - typically to merge some branches together.
```

guides/model_creation/ComplexModels.livemd
```diff
@@ -0,0 +1,84 @@
...
+```
+
+```elixir
+params = init_fn.(template, %{})
+predict_fn.(params, Nx.iota({2, 8}, type: :f32))
+```
+
+As your architectures grow in complexity, you might find yourself reaching for better abstractions to organize your model creation code. For example, PyTorch models are often organized into `nn.Module`. The equivalent of an `nn.Module` in Axon is a regular Elixir function. If you're translating models from PyTorch to Axon, it's natural to create one Elixir function per `nn.Module`.
+
+You should write your models as you would write any other Elixir code---you don't need to worry about any framework specific constructs:
```
jonatanklosko, line 48:
```suggestion
You should write your models as you would write any other Elixir code - you don't need to worry about any framework specific constructs:
```

jonatanklosko (approved):
Beautiful!

josevalim (approved):

PR #361, opened 2022-09-16
chore: update torchx and unskip more tests

polvalente:


lib/axon/compiler.ex
```diff
@@ -145,7 +145,17 @@ defmodule Axon.Compiler do
...
-    tensor = Nx.backend_copy(tensor, Nx.Defn.Expr)
+    tensor =
+      tensor =
+      case tensor do
+        %Nx.Tensor{data: %Nx.Defn.Expr{}} ->
+          tensor
+
+        t ->
+          Nx.backend_copy(t, Nx.Defn.Expr)
+      end
```
josevalim, line 156: We should fix this in Nx. A backend copy of Nx.Defn.Expr to Nx.Defn.Expr should return itself.

lib/axon/layers.ex
```diff
@@ -105,6 +105,29 @@ defmodule Axon.Layers do
...
+          case k do
+            %Nx.Tensor{data: %Nx.Defn.Expr{}} ->
+              k
+
+            t ->
+              Nx.backend_copy(t, Nx.Defn.Expr)
+          end
+
+        {i, k}
+      end)
```
josevalim, line 129: We should revert this. According to `defn`, this will always be tensors. If it isn't, it is because of something upstream.

lib/axon/losses.ex
```diff
@@ -1032,9 +1032,12 @@ defmodule Axon.Losses do
...
 
-        {loss_b, _, _} =
-          while {loss_b = 0.0, s = st_lims[0][0], s_pred0}, s <= st_lims[0][1] do
-            {Nx.add(loss_b, Nx.exp(s_pred0[s])), s + 1, s_pred0}
+        s = st_lims[0][0]
+
+        {loss_b, _, _, _} =
+          while {loss_b = 0.0, xyz = s, abc = s_pred0, abobrinha = st_lims},
+                xyz <= abobrinha[0][1] do
+            {Nx.add(loss_b, Nx.exp(abc[xyz])), xyz + 1, abc, abobrinha}
```
josevalim, line 1040: We should remove abobrinha. :D

lib/axon/compiler.ex
```diff
@@ -144,7 +144,6 @@ defmodule Axon.Compiler do
          _
        ) do
     op_counts = Map.update(op_counts, :constant, 1, fn x -> x + 1 end)
-
     tensor = Nx.backend_copy(tensor, Nx.Defn.Expr)
```
josevalim, line 147:
```suggestion

    tensor = Nx.backend_copy(tensor, Nx.Defn.Expr)
```

mix.exs
```diff
@@ -36,7 +36,7 @@ defmodule Axon.MixProject do
   defp deps do
     [
       {:exla, "~> 0.3.0", [only: :test] ++ exla_opts()},
-      {:torchx, "~> 0.3.0", [only: :test] ++ torchx_opts()},
+      #{:torchx, "~> 0.3.0", [only: :test] ++ torchx_opts()},
```
josevalim, line 39:
```suggestion
      {:torchx, "~> 0.3.0", [only: :test] ++ torchx_opts()},
```

lib/axon/updates.ex
```diff
@@ -756,13 +756,15 @@ defmodule Axon.Updates do
 
   def add_decayed_weights({init_fn, apply_fn} = combinator, opts)
       when is_function(init_fn, 1) and is_function(apply_fn, 3) and is_list(opts) do
-    stateless(combinator, &apply_weight_decay(&1, &2, opts))
+    stateless(combinator, fn updates, params ->
+      opts = Nx.Defn.Kernel.keyword!(opts, decay: 0.0)
+      # Decay can be a tensor, that's why we preprocess it before-hand
+      # and pass it as argument to defn instead of as an optioh.
```
jonatanklosko, line 762:
```suggestion
      # and pass it as argument to defn instead of as an option.
```

josevalim:
:green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

PR #364, opened 2022-09-20
Show inputs without edges in the graph visualization

jonatanklosko:
Currently a lonely input is not shown in the visualization, but it is actually useful:

image

Separating nodes and edges also has the advantages of reducing the graph text size if the same node has multiple edges.

I changed inputs to use a distinctive parallelogram (it seems that's used for inputs in flowcharts):

image


jonatanklosko:
Here's an alternative formatting I also like:

image

seanmor5 (approved):

PR #366, opened 2022-09-20
Update notebooks

jonatanklosko:
A first batch of changes

  * some tiny adjustments to the guides
  * updated MNIST example to match Axon changes
  * rewrote the Horses & Humans notebook, so that it doesn't require downloading files manually, use less modules

My favourite bit:

[horses_or_humans.webm](https://user-images.githubusercontent.com/17034772/191363790-b6d3fd3c-4b7f-478d-85b6-811f12306898.webm)


notebooks/vision/humans_or_horses.livemd
```diff
@@ -1,258 +0,0 @@
...
-
-```elixir
-HorsesHumans.ModelTraining.run()
-```
-
-## Conclusions
-
-We have an accuracy of almost 1.0.  That is great right?  Hmmm. Not so great.  With this notebook we've demonstrated the ability for a neural net model to memorize a dataset.  The model knows how to distinguish each image with the kind of image it is.  Training a model is tricky.  It may learn unintended characteristics.  Do all of the horses have something in the background that the model can make it's decision on?
-
-For useful models, we need to compare the model against images that it hasn't seen.  Comparison against unseen input data measures the generalization of the model with respect to the unseen images.  Using the model against further unseen images will help with measure the bias of the model.  How well will this model perform against a real horse or pony?
```
jonatanklosko, line 258: I removed this, I believe this was actually a bug. The target class was assigned based on `String.contains?(filename, "horses")`, checking against the full path, and I think the path would always include `"horses"` as in `"data/humans_or_horses/{horses,humans}/*"`. So we would train the model on a single class :)

guides/model_creation/model_hooks.livemd
```diff
@@ -370,7 +370,7 @@ Nx.Defn.grad(fn params -> predict_fn.(params, input) end).(params)
 }
 ```
 
-Finally, you can specify hooks to only run when the model is built in a certain mode such as training and inference mode. You can read more about training an inference mode in [Training and Inference Mode](guides/model_execution/training_and_inference_mode.livemd):
+Finally, you can specify hooks to only run when the model is built in a certain mode such as training and inference mode. You can read more about training an inference mode in [Training and inference mode](guides/model_execution/training_and_inference_mode.livemd):
```
seanmor5, line 373:
```suggestion
Finally, you can specify hooks to only run when the model is built in a certain mode such as training and inference mode. You can read more about training and inference mode in [Training and inference mode](guides/model_execution/training_and_inference_mode.livemd):
```

notebooks/vision/horses_or_humans.livemd
```diff
@@ -0,0 +1,253 @@
...
+end
+```
+
+## Building the model
+
+The next step is creating our model. In this notebook, we choose the classic Convolutional Neural Network architecture. Let's dive in to the core components of a CNN.
+
+<!-- livebook:{"break_markdown":true} -->
+
+`Axon.conv/3` adds a convolutional layer, which is at the core of a CNN. A convolutional layer applies a filter function throughout the image, sliding a window with shape `:kernel_size`. As opposed to dense layers, a convolutional layer captures spacial information, because it works across multiple axes. This feature is a natural fit for images.
```
seanmor5, line 145:
```suggestion
`Axon.conv/3` adds a convolutional layer, which is at the core of a CNN. A convolutional layer applies a filter function throughout the image, sliding a window with shape `:kernel_size`. As opposed to dense layers, a convolutional layer exploits weight sharing to better model data where locality matters. This feature is a natural fit for images.
```

notebooks/vision/horses_or_humans.livemd
```diff
@@ -0,0 +1,253 @@
...
+
+<!-- livebook:{"break_markdown":true} -->
+
+| ![](https://miro.medium.com/max/1400/1*KkqxjvXTIV_b365B41ltfg.png)    |
+| :-------------------------------------------------------------------: |
+| Figure 3: The difference between standard dropout and spatial dropout |
+
+<!-- livebook:{"break_markdown":true} -->
+
+Knowing the relevant building blocks, let's build our network! It will have a convolutional part, composed of convolutional and pooling layers, this part should capture the spacial features of an image. Then at the end, we will add a dense layer with 512 neurons fed with all the spacial features, and a final two-neuron layer for as our classification output.
```
seanmor5, line 175:
```suggestion
Knowing the relevant building blocks, let's build our network! It will have a convolutional part, composed of convolutional and pooling layers, this part should capture the spatial features of an image. Then at the end, we will add a dense layer with 512 neurons fed with all the spatial features, and a final two-neuron layer for as our classification output.
```

seanmor5 (approved):
Beautiful work! @jonatanklosko I just added a couple comments for typos/writing edits but overall this is great!

PR #373, opened 2022-09-21
Update notebooks

jonatanklosko:
I broke down the XOR example into individual cells, rather than a single module. Also added this visualization:

image


jonatanklosko:
Okay, I went through and rewrote the other example notebooks in the same spirit, we should be good :)

notebooks/basics/xor.livemd
```diff
@@ -0,0 +1,117 @@
...
+])
+
+Nx.Defn.default_options(compiler: EXLA)
+
+alias VegaLite, as: Vl
+```
+
+## Introduction
+
+In this notebook we try to create a model and learn it the **logical XOR**.
```
seanmor5, line 18:
```suggestion
In this notebook we will try to create a model and teach it the **logical XOR**.
```

notebooks/basics/xor.livemd
```diff
@@ -0,0 +1,117 @@
...
+
+Even though XOR seems like a trivial operation, it cannot be modeled using a single dense layer ([single-layer perceptron](https://en.wikipedia.org/wiki/Feedforward_neural_network#Single-layer_perceptron)). The underlying reason is that the classes in XOR are not linearly separable. We cannot draw a straight line to separate the points $(0,0)$, $(1,1)$ from the points $(0,1)$, $(1,0)$. To model this properly we need to turn to deep learning methods. They can separate data with much more complicated patterns within the data.
```
seanmor5, line 20:
```suggestion
Even though XOR seems like a trivial operation, it cannot be modeled using a single dense layer ([single-layer perceptron](https://en.wikipedia.org/wiki/Feedforward_neural_network#Single-layer_perceptron)). The underlying reason is that the classes in XOR are not linearly separable. We cannot draw a straight line to separate the points $(0,0)$, $(1,1)$ from the points $(0,1)$, $(1,0)$. To model this properly, we need to turn to deep learning methods. They can separate data with much more complicated patterns within the data.
```

notebooks/basics/xor.livemd
```diff
@@ -0,0 +1,117 @@
...
+
+## The model
+
+Let's start with the model. We need two inputs, since XOR has two operands, we then concatenate them into a single input vector with `Axon.concatenate/3`. Then we have one hidden layer and one output layer, both of them dense.
```
seanmor5, line 24:
```suggestion
Let's start with the model. We need two inputs, since XOR has two operands. We then concatenate them into a single input vector with `Axon.concatenate/3`. Then we have one hidden layer and one output layer, both of them dense.
```

notebooks/basics/xor.livemd
```diff
@@ -0,0 +1,117 @@
...
+  Stream.repeatedly(fn ->
+    x1 = Nx.random_uniform({batch_size, 1}, 0, 2)
+    x2 = Nx.random_uniform({batch_size, 1}, 0, 2)
+    y = Nx.logical_xor(x1, x2)
+
+    {%{"x1" => x1, "x2" => x2}, y}
+  end)
+```
+
+Here's how a sample batch looks like:
```
seanmor5, line 58:
```suggestion
Here's how a sample batch looks:
```

notebooks/basics/xor.livemd
```diff
@@ -0,0 +1,117 @@
...
+
+params =
+  model
+  |> Axon.Loop.trainer(:binary_cross_entropy, :sgd)
+  |> Axon.Loop.run(data, %{}, epochs: epochs, iterations: 1000)
+```
+
+## Trying the model
+
+Finally, we can test our model for sample data.
```
seanmor5, line 79:
```suggestion
Finally, we can test our model on sample data.
```

notebooks/basics/xor.livemd
```diff
@@ -0,0 +1,117 @@
...
+```elixir
+Axon.predict(model, params, %{
+  "x1" => Nx.tensor([[0]]),
+  "x2" => Nx.tensor([[1]])
+})
+```
+
+Try other combinations of $x_1$ and $x_2$ and see what the output is. To improve the model performance, you can increase the number of training epochs.
+
+## Visualizing the model
```
seanmor5, line 90:
```suggestion
## Visualizing the model predictions
```

notebooks/basics/xor.livemd
```diff
@@ -0,0 +1,117 @@
...
+
+The original XOR we modeled only works with binary values $0$ and $1$, however our model operates in continuous space. This means that we can give it $x_1 = 0.5$, $x_2 = 0.5$ as input and we expect _some_ output. We can use this
```
seanmor5, line 92:
```suggestion
The original XOR we modeled only works with binary values $0$ and $1$, however our model operates in continuous space. This means that we can give it $x_1 = 0.5$, $x_2 = 0.5$ as input and we expect _some_ output. We can use this to visualize the non-linear relationship between inputs `x1`, `x2` and outputs that our model has learned.
```

notebooks/generative/fashionmnist_autoencoder.livemd
```diff
@@ -73,142 +63,150 @@ Next, we pass the output of the encoder to the decoder and try to reconstruct th
...
+And a different image:
+
+```elixir
+{
+  mean_square_error.(shoe_image, other_image),
+  mean_absolute_erorr.(shoe_image, other_image)
+}
+```
+
+As we can see, the noised image has a non-zero MSE and MAE but is much smaller than the error of two completely different pictures. In other words, both of these error types measure the level of similarity between images. A small error implies decent prediction values. On the other hand, a large error value suggests poor quality of predictions.
```
seanmor5, line 191: You don't have to include this, just a fun fact. MAE and MSE can also be called the L1 and L2 loss respectively for the L1 and L2 norm. The L2 loss (MSE) is typically preferred because it's a smoother function whereas L1 is often difficult to optimize with SGD.

notebooks/structured/credit_card_fraud.livemd
```diff
@@ -1,141 +1,176 @@
...
+    |> Nx.stack(axis: 1)
   end
 
-defp normalize_data(df) do
-  df
-  |> Explorer.DataFrame.names()
-  |> Map.new(&{&1, normalize(&1)})
-  |> then(&Explorer.DataFrame.mutate(df, &1))
-end
+  defn normalize_features(tensor) do
```
seanmor5, line 67: Perhaps we can use Scholar here and demo Axon+Scholar in tandem?

seanmor5 (approved):
Amazing work as always @jonatanklosko! I just left a few minor suggestions

notebooks/structured/credit_card_fraud.livemd
jonatanklosko: Definitely, I would wait for a Scholar release though. The other functions look common enough to also have _somewhere_ eventually.

notebooks/generative/fashionmnist_autoencoder.livemd
jonatanklosko: Worth including, thanks! I also changed the MAE and MSE definitions, because I think they were not adjusted for the multidimensional images. I changed them to work on a single example for simplicity. Please check if it makes sense (4aa162096e4562b11fc280282bc90ca98c1840cb) and feel free to merge then :)

PR #374, opened 2022-09-23
Add an example notebook for a variational autoencoder

hanrelan:
This is still a work in progress, I haven't gotten to the variational part yet.

But I'm finding that the training on Fashion MNIST results in getting `:nan` parameters after ~10-15 epochs. I *think* this is due to exploding gradients, but I'm not sure how to debug it or apply gradient clipping in Axon.

### DIFF
--- a/notebooks/generative/fashionmnist_vae.livemd
+++ b/notebooks/generative/fashionmnist_vae.livemd
@@ -14,6 +14,11 @@ Mix.install([
 ])
 
 alias VegaLite, as: Vl
+
+# This speeds up all our `Nx` operations without having to use `defn`
+Nx.global_default_backend(EXLA.Backend)
+
+:ok
 ```
 
 ## Introduction
@@ -66,11 +71,13 @@ defmodule Data do
   end
 
   @doc """
-  Converts a tensor of images into paired, batched images for model training
+  Converts a tensor of images into random batches of paired images for model training
   """
   def prepare_training_data(images, batch_size) do
-    batched_images = images |> Nx.to_batched(batch_size)
-    Stream.zip(batched_images, batched_images)
+    Stream.flat_map([nil], fn nil ->
+      images |> Nx.shuffle(axis: :images) |> Nx.to_batched(batch_size)
+    end)
+    |> Stream.map(fn batch -> {batch, batch} end)
   end
 end
 ```
--- a/notebooks/generative/fashionmnist_vae.livemd
+++ b/notebooks/generative/fashionmnist_vae.livemd
@@ -8,31 +8,36 @@ Mix.install([
   {:req, "~> 0.3.1"},
   {:kino, "~> 0.6.2"},
   {:scidata, "~> 0.1.9"},
-  {:stb_image, "~> 0.5.2"}
+  {:stb_image, "~> 0.5.2"},
+  {:kino_vega_lite, "~> 0.1.3"},
+  {:vega_lite, "~> 0.1.6"}
 ])
+
+alias VegaLite, as: Vl
 ```
 
 ## Introduction
 
-In this notebook, we'll be building a variational autoencoder (VAE). This will help demonstrate splitting up models, defining custom layers in Axon and a few additional Kino tricks for training models.
+In this notebook, we'll be building a variational autoencoder (VAE). This will help demonstrate splitting up models, defining custom layers and loss functions, using multiple outputs, and a few additional Kino tricks for training models.
 
-This notebook builds on the [denoising autoencoder example](https://github.com/elixir-nx/axon/blob/main/notebooks/generative/mnist_autoencoder_using_kino.livemd) and turns the simple autoencoder into a variational one for the same dataset.
+This notebook builds on the [denoising autoencoder example](mnist_autoencoder_using_kino.livemd) and turns the simple autoencoder into a variational one for the same dataset.
 
 ## Training a simple autoencoder
 
-This section will proceed without much explanation as most of it is extracted from [denoising autoencoder example](https://github.com/elixir-nx/axon/blob/main/notebooks/generative/mnist_autoencoder_using_kino.livemd). If anything here doesn't make sense, take a look at that notebook for an explanation.
+This section will proceed without much explanation as most of it is extracted from [denoising autoencoder example](mnist_autoencoder_using_kino.livemd). If anything here doesn't make sense, take a look at that notebook for an explanation.
 
 ```elixir
-defmodule Utils do
+defmodule Data do
   @moduledoc """
-  A module to hold useful utilities, mostly extracted from the previous notebook
+  A module to hold useful data processing utilities, 
+  mostly extracted from the previous notebook
   """
 
   @doc """
-    Converts the given image into a `Kino.Image`.
+  Converts the given image into a `Kino.Image`.
 
-    `image` must be a single channel `Nx` tensor with pixel values between 0 and 1.
-    `height` and `width` are the output size in pixels
+  `image` must be a single channel `Nx` tensor with pixel values between 0 and 1.
+  `height` and `width` are the output size in pixels
   """
   def image_to_kino(image, height \\ 200, width \\ 200) do
     image
@@ -46,7 +51,7 @@ defmodule Utils do
   end
 
   @doc """
-    Converts image data from `Scidata.MNIST` into an `Nx` tensor and normalizes it.
+  Converts image data from `Scidata.MNIST` into an `Nx` tensor and normalizes it.
   """
   def preprocess_data(data) do
     {image_data, _labels} = data
@@ -61,7 +66,7 @@ defmodule Utils do
   end
 
   @doc """
-    Converts a tensor of images into paired, batched images for model training
+  Converts a tensor of images into paired, batched images for model training
   """
   def prepare_training_data(images, batch_size) do
     batched_images = images |> Nx.to_batched(batch_size)
@@ -71,18 +76,36 @@ end
 ```
 
 ```elixir
-train_images = Utils.preprocess_data(Scidata.FashionMNIST.download())
-test_images = Utils.preprocess_data(Scidata.FashionMNIST.download_test())
+train_images = Data.preprocess_data(Scidata.FashionMNIST.download())
+test_images = Data.preprocess_data(Scidata.FashionMNIST.download_test())
 
-Kino.render(train_images[[images: 0]] |> Utils.image_to_kino())
-Kino.render(test_images[[images: 0]] |> Utils.image_to_kino())
+Kino.render(train_images[[images: 0]] |> Data.image_to_kino())
+Kino.render(test_images[[images: 0]] |> Data.image_to_kino())
 
 :ok
 ```
 
-Our simple autoencoder model. We won't be using a denoising autoencoder here.
+Now for our simple autoencoder model. We won't be using a denoising autoencoder here.
 
 Note that we're giving each of the layers a name - the reason for this will be apparent later.
+
+I'm also using a small custom layer to shift and scale the output of the sigmoid layer slightly so it can hit the 0 and 1 targets. I noticed the gradients tend to explode without this.
+
+```elixir
+defmodule CustomLayer do
+  import Nx.Defn
+
+  def scaling_layer(%Axon{} = input, _opts \\ []) do
+    Axon.layer(&scaling_layer_impl/2, [input])
+  end
+
+  defnp scaling_layer_impl(x, _opts \\ []) do
+    x
+    |> Nx.subtract(0.05)
+    |> Nx.multiply(1.2)
+  end
+end
+```
 
 ```elixir
 model =
@@ -100,6 +123,7 @@ model =
   |> Axon.dense(128, activation: :relu, name: "decoder_layer_2")
   |> Axon.dense(256, activation: :relu, name: "decoder_layer_3")
   |> Axon.dense(784, activation: :sigmoid, name: "decoder_layer_4")
+  |> CustomLayer.scaling_layer()
   # Turn it back into a 28x28 single channel image
   |> Axon.reshape({1, 28, 28})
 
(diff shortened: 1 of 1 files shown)


notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,304 @@
...
+  {:scidata, "~> 0.1.9"},
+  {:stb_image, "~> 0.5.2"}
+])
+```
+
+## Introduction
+
+In this notebook, we'll be building a variational autoencoder (VAE). This will help demonstrate splitting up models, defining custom layers in Axon and a few additional Kino tricks for training models.
+
+This notebook builds on the [denoising autoencoder example](https://github.com/elixir-nx/axon/blob/main/notebooks/generative/mnist_autoencoder_using_kino.livemd) and turns the simple autoencoder into a variational one for the same dataset.
```
seanmor5, line 19: I think the relative path works:

```suggestion
This notebook builds on the [denoising autoencoder example](mnist_autoencoder_using_kino.livemd) and turns the simple autoencoder into a variational one for the same dataset.
```

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,304 @@
...
+
+## Training a simple autoencoder
+
+This section will proceed without much explanation as most of it is extracted from [denoising autoencoder example](https://github.com/elixir-nx/axon/blob/main/notebooks/generative/mnist_autoencoder_using_kino.livemd). If anything here doesn't make sense, take a look at that notebook for an explanation.
```
seanmor5, line 23:
```suggestion
This section will proceed without much explanation as most of it is extracted from [denoising autoencoder example](mnist_autoencoder_using_kino.livemd). If anything here doesn't make sense, take a look at that notebook for an explanation.
```

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,304 @@
...
+
+```elixir
+defmodule Utils do
```
seanmor5, line 26: Just a nit, but this all seems related to data, so maybe a better name is `defmodule Data` or something...

```suggestion
defmodule Data do
```

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,304 @@
...
+    |> Nx.as_type(:u8)
+    |> Nx.transpose(axes: [:height, :width, :channels])
+    |> StbImage.from_nx()
+    |> StbImage.resize(height, width)
+    |> StbImage.to_binary(:png)
+    |> Kino.Image.new(:png)
+  end
+
+  @doc """
+    Converts image data from `Scidata.MNIST` into an `Nx` tensor and normalizes it.
```
seanmor5, line 49:
```suggestion
  Converts image data from `Scidata.MNIST` into an `Nx` tensor and normalizes it.
```

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,304 @@
...
+  @moduledoc """
+  A module to hold useful utilities, mostly extracted from the previous notebook
+  """
+
+  @doc """
+    Converts the given image into a `Kino.Image`.
+
+    `image` must be a single channel `Nx` tensor with pixel values between 0 and 1.
+    `height` and `width` are the output size in pixels
+  """
```
seanmor5, line 36:
```suggestion
  @doc """
  Converts the given image into a `Kino.Image`.

  `image` must be a single channel `Nx` tensor with pixel values between 0 and 1.
  `height` and `width` are the output size in pixels
  """
```

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,304 @@
...
+    images_binary
+    |> Nx.from_binary(type)
+    # Since pixels are organized row-wise, reshape into rows x columns
+    |> Nx.reshape(shape, names: [:images, :channels, :height, :width])
+    # Normalize the pixel values to be between 0 and 1
+    |> Nx.divide(255)
+  end
+
+  @doc """
+    Converts a tensor of images into paired, batched images for model training
```
seanmor5, line 64: No need to indent docstrings :)

```suggestion
  Converts a tensor of images into paired, batched images for model training
```

jonatanklosko:
> I think this is due to exploding gradients

You can try [this](https://github.com/elixir-nx/axon/blob/d8dfdaf31623bd313eccb0b4bcb80e334c38beeb/notebooks/vision/horses_or_humans.livemd#extra-gradient-centralization) (just guessing though :D).

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,304 @@
...
+:ok
+```
+
+When training, it can be useful to stop execution early - either when you see it's failing and you don't want to waste time waiting for the remaining epochs to finish, or if it's good enough and you want to start experimenting with it.
+
+The `livebook_early_stop/1` function below is a handy handler to give us a `Kino.Control.button` that will stop the training loop when clicked.
+
+```elixir
+defmodule LivebookTraining do
+  def livebook_early_stop(loop) do
```
seanmor5, line 130: This is **really** cool and definitely belongs in a library somewhere. If you want to open a PR to Axon I would gladly accept it, but we are also planning on a Kino-Axon specific library, so perhaps it's best that lives there. Either way I think opening a PR in Axon is the best thing to do right now so we remember to extract this out when Kino-Axon integration comes to life :)

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,304 @@
...
+  params = state.step_state[:model_state]
+  image_index = Enum.random(0..(Nx.axis_size(test_images, :images) - 1))
+  image = combined_input_output.(params, image_index) |> Utils.image_to_kino(200, 400)
+  Kino.Frame.append(frame, image)
+  {:continue, state}
+end
+
+params =
+  model
+  |> Axon.Loop.trainer(:mean_squared_error, Axon.Optimizers.adam(0.001))
```
seanmor5, line 186: This is the loop that NaNs? I can't see any obvious reason why. A few things you can try:

1. Play with the `eps` option in `adam`...it's a numerical stability term, I don't think it should matter but it might here
2. Lower the learning rate and see if it still learns without collapsing to NaN. Though 1e-3 still seems small enough for this kind of problem.
3. Try BCE loss. I've seen it used successfully in Autoencoders before, but again I don't see why this would lead to NaNs

For gradient clipping, there are a few options. If you take a look at Axon's optimizers, you'll see they're all actually rather simple. For example, I think `adam` is actually just:

```elixir
Axon.Updates.scale_by_adam()
|> Axon.Updates.scale()
```

You can implement a custom one and add one of the clipping functions:

```elixir
Axon.Updates.clip_by_norm()
|> Axon.Updates.scale_by_adam()
|> Axon.Updates.scale()
```

Or you can compose existing optimizers with updates like this:

```elixir
optimizer = Axon.Updates.compose(Axon.Updates.clip_by_norm(), Axon.Optimizers.adam(1.0e-3))
```

To inspect the gradients of each layer, you can attach backward hooks to the model with `Axon.attach_hook/3`. If you want to apply one to every layer of the model without needing to do so manually, you can try:

```elixir
instrumented_model = Axon.map_nodes(model, fn layer ->
  Axon.attach_hook(layer, &IO.inspect/1, on: :backward)
end)
```

For something more robust you can probably also visualize the gradients and the results of the forward pass of intermediate layers too by attaching to the `on: :forward` event

seanmor5 (commented):
Hi @hanrelan this looks great so far! I left some comments up to the original Autoencoder training, I'm guessing that's where you were getting NaNs? Great work so far!

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,304 @@
...
+
+    Task.start_link(fn ->
+      Kino.Control.subscribe(stop_button, :stop)
+
+      receive do
+        {:stop, _info} ->
+          Kino.Control.unsubscribe(stop_button)
+          Agent.update(button_agent, fn _ -> :stop end)
+      end
+    end)
```
jonatanklosko, line 145:
```suggestion
    Kino.Control.subscribe(stop_button, fn _event ->
      Agent.update(button_agent, fn _ -> :stop end)
    end)
```

This should do the job :D It runs asynchronously, and since we remove the button from the frame, the handler is unsubscribed automatically.

---

Sidenote: we will have more flexible APIs for dealing with events/streams in the next kino release. To asynchronously consume just one click we would do:

```elixir
stop_button
|> Kino.Control.stream()
|> Stream.take(1)
|> Kino.listen(fn _event ->
  Agent.update(button_agent, fn _ -> :stop end)
end)
```

notebooks/generative/fashionmnist_vae.livemd
hanrelan: I was thinking the same thing - a `Kino-Axon` library would be the perfect place for it. I'll create a PR in the meantime, as soon as I figure out how to write a test for it :)

seanmor5: Don't let the test hold you back from a PR, we can figure something out during the review!

hanrelan:
I figured out the problem and got it to train. Turns out the model was failing because it couldn't produce the target (0 and 1) with the sigmoid, so adding some rescaling stabilized it.

All the other suggestions were incorporated.

@seanmor5 I'll create a separate pull request for adding the `AxonKino` stuff to Axon. I'll add it to `Axon.Loop` if that makes sense.

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,526 @@
...
+    # Normalize the pixel values to be between 0 and 1
+    |> Nx.divide(255)
+  end
+
+  @doc """
+  Converts a tensor of images into paired, batched images for model training
+  """
+  def prepare_training_data(images, batch_size) do
+    batched_images = images |> Nx.to_batched(batch_size)
+    Stream.zip(batched_images, batched_images)
```
hanrelan, line 73: @seanmor5 @jonatanklosko Is there a way to make these batches random? Doing `images |> Nx.shuffle |> Nx.to_batched(batch_size)` doesn't do the right thing since that only shuffles it once. I want the shuffling to be different for every epoch? I couldn't figure out how to do this - I think it probably has something to do with `Stream.resource/3` but couldn't figure out how to make that work without a `Stream.next` function.

I think it'd be useful if `Nx.to_batched` took a `randomize` option

hanrelan: I came up with this but it seems a little messy:

```elixir
shuffled_batches = Stream.resource(
  fn -> 
    batches = Nx.iota({100, 1}) |> Nx.shuffle |> Nx.to_batched(5)
    {0, batches}
  end,
  fn {i, batches} -> 
    case Enum.at(batches, i) do
      nil -> {:halt, {i, batches}}
      value -> {[value], {i+1, batches}}
    end
  end,
  fn _acc -> nil end
)
data = Stream.map(shuffled_batches, fn batch -> {batch, batch} end)
```

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,526 @@
...
+    |> CustomLayer.scaling_layer()
+    # Turn it back into a 28x28 single channel image
+    |> Axon.reshape({1, 28, 28})
+  end
+
+  def autoencoder() do
+    encoder_partial = encoder_partial()
+    encoder = encoder()
+    autoencoder = decoder(encoder)
+    Axon.container(%{mu_sigma: encoder_partial, reconstruction: autoencoder})
```
hanrelan, line 431: Not sure if there was a better way to do this...

seanmor5: This is pretty much exactly how I would do it too :)

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,526 @@
...
+    #   |> Vl.mark(:point, tooltip: true)
+    #   |> Vl.encode_field(:x, "epoch", type: :ordinal)
+    #   |> Vl.encode_field(:y, "loss", type: :quantitative),
+    #   Vl.new()
+    #   |> Vl.mark(:point, tooltip: true)
+    #   |> Vl.encode_field(:x, "epoch", type: :ordinal)
+    #   |> Vl.encode_field(:y, "val_loss", type: :quantitative)
+    # ])
+    # |> Kino.VegaLite.new()
+    # |> Kino.render()
```
seanmor5, line 214: Were you having issues plotting val loss here?

notebooks/generative/fashionmnist_vae.livemd
seanmor5: You could use `Enum.shuffle` but I think that also would only shuffle it once. I can probably add something to `Axon.Loop` that shuffles the enumerated dataset every epoch

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,526 @@
...
+  import Nx.Defn
+
+  def scaling_layer(%Axon{} = input, _opts \\ []) do
+    Axon.layer(&scaling_layer_impl/2, [input])
+  end
+
+  defnp scaling_layer_impl(x, _opts \\ []) do
+    x
+    |> Nx.subtract(0.05)
+    |> Nx.multiply(1.2)
```
seanmor5, line 105: This is interesting, so this got rid of NaNs...where did these values come from?

notebooks/generative/fashionmnist_vae.livemd
```diff
@@ -0,0 +1,526 @@
...
+    #   |> Vl.encode_field(:x, "epoch", type: :ordinal)
+    #   |> Vl.encode_field(:y, "val_loss", type: :quantitative)
+    # ])
+    # |> Kino.VegaLite.new()
+    # |> Kino.render()
+
+    handler = fn state ->
+      %Axon.Loop.State{metrics: metrics, epoch: epoch} = state
+      loss = metrics["loss"] |> Nx.to_number()
+      val_loss = metrics["validation_0"]["loss"] |> Nx.to_number()
```
seanmor5, line 219: Looking at this we should probably make the validation metrics show up as a nested map like:

```
%{"validation" => %{0 => %{."loss" => ...}}}
```

Do you think that would be more intuitive to work with?

seanmor5 (commented):
Really cool stuff @hanrelan ! Thank you for putting this together. This is definitely the most involved example contribution we've gotten so far, was there anything you felt was unintuitive or clunky when implementing? Any suggestions for making the APIs easier to work with?

notebooks/generative/fashionmnist_vae.livemd
jonatanklosko: @hanrelan we need a stream where all images are shuffled on every run, so I'm thinking this:

```elixir
Stream.flat_map([images], fn images ->
  images |> Nx.shuffle() |> Nx.to_batched(batch_size)
end)
```

In fact, it could be `Stream.flat_map([:whatever], fn _ -> ... end)` :)

notebooks/generative/fashionmnist_vae.livemd
hanrelan: Oops I forgot I left this in - I'll delete it. No issues plotting val loss, I wasn't using VegaLite correctly (didn't need to use multiple layers)

notebooks/generative/fashionmnist_vae.livemd
hanrelan: These values were just made up to get the sigmoid to cover 0 and 1 (I think this is a similar trick to what fast.ai uses). Based on my understanding of `layer_norm` I don't think it'll enable the sigmoid to reach 1.0 as an output, is that right?

The issue seemed to be that since the mnist values are often actually 0 and 1 and sigmoid couldn't get there, as the network converged it droves certain params to more extremes to try to reach those values.

notebooks/generative/fashionmnist_vae.livemd
hanrelan: I think my main confusion was what the `0` represented. When I first wrote it I assumed it represented the epoch. Either one seems fine, your alternative seems a little cleaner

hanrelan:
Thanks for the suggestions - I'll clean up this PR and get it ready for merging.

I've been thinking about the APIs a lot. It's tricky because it's a different paradigm from pytorch (which is what I'm used to), so it's not clear to me what was difficult because I just wasn't used to it vs because the APIs could be better.

A couple things I noticed:
- One of the more painful bits is actually livebook requiring the cells be executed linearly. I get the benefit, but it does make it harder for exploration and machine learning workflows. eg. restarting the entire notebook when I wanted to add VegaLite was painful
- I'm not sure how this would work but one thing I like about pytorch is that there is no separation between layers and models. A layer can be used as a model, a model can be used as a layer. I think after putting this together I finally figured out how to use layers/models interchangeably here (by turning them into functions that take an input) but I wonder if layers and models should all be functions to start with? I'm not sure how that plays out but I might do a little exploration of that space.
- A bug I noticed - `Axon.Display.as_table` doesn't know how to deal with `container` outputs
- Dataloading def seemed trickier than pytorch. I'll try Jonatan's suggestion on shuffling but it definitely feels non obvious how to do some things. Also it required diving into the source to understand how `Axon` uses the data stream it's given (to figure out if shuffling the stream would do the right thing). Makes me wonder if `Axon` should provide some helpers around data loading as well to make it easier for folks.

notebooks/generative/fashionmnist_vae.livemd
hanrelan: Hmm I tried this @jonatanklosko and it became really slow to shuffle. If I tried putting it into a `defn` it complains that `to_batched` isn't allowed within `defn`. Will just leave it unshuffled for now and leave this for another day :)

hanrelan:
Cleaned it up a bit, I think it's ready to merge now unless you notice anything else funky

notebooks/generative/fashionmnist_vae.livemd
jonatanklosko: Ah, you are only calling EXLA compiler explicitly in the notebook, so regular operations are done using the binary backend. Try adding `Nx.global_default_backend(EXLA.Backend)` below mix install and see if that speeds it up.

hanrelan: That worked! I also needed to specify `axis: :images` on the `Nx.shuffle` so it didn't shuffle everything. Specifying the axis sped it up to about a second, specifying the global backend made it feel instant after that.

PR #380, opened 2022-10-03
Update initializers API to split keys in defn

### LINKED ISSUE #379: Optimize initialization with keys
Right now initialization is super slow because of all of the key splitting stuff taking place outside of `defn`, we should fix this. For large models `init` is slower than `predict` which is unacceptable

seanmor5:
Resolves #379


lib/axon/initializers.ex
```diff
@@ -47,7 +47,8 @@ defmodule Axon.Initializers do
   ## Examples
 
       iex> init_fn = Axon.Initializers.zeros()
-      iex> init_fn.({2, 2}, {:f, 32}, Nx.Random.key(1))
+      iex> {_key, out} = init_fn.({2, 2}, {:f, 32}, Nx.Random.key(1))
```
jonatanklosko, line 50: Wouldn't `{out, key}` be more natural, given that `out` is the primary output and we also pass seed as the last argument? (:see_no_evil: :see_no_evil: :see_no_evil:)

seanmor5: Yeah I didn't want to make the initializers API any different than Nx, I wrote it the way you noted initially but then all of the tests failed because I assumed Nx returned the output first and key last

josevalim: @seanmor5 let's change Nx then?

josevalim: I just changed Nx to return `{out, key}`.

PR #381, opened 2022-10-04
Fix Axon.Loop.reduce_lr_on_plateau documentation

NduatiK:
Fix Axon.Loop.reduce_lr_on_plateau documentation
1. [`Axon.Loop.validate/3`](https://hexdocs.pm/axon/Axon.Loop.html#validate/3) expects the model as the first parameter
2. The accuracy metric is named "accuracy" by default, `validate` reuses the name.


seanmor5:
Thank you!

PR #382, opened 2022-10-12
Return tuple of loop state and transformed state

### LINKED ISSUE #362: Default loop output transforms are too intrusive
For supervised training loops, we provide a convenience output transform which ensures only the model state is returned from the training loop. This means you always lose the entire training state, which might be of interest later on.

I propose instead that we return a tuple:

`{loop_state, transformed_state}` which always returns the whole state, as well as a transformed version. That way you never accidentally lose the entire state.

Ian-GL:
If I understood correctly #362 , the new output of the transform should be `{%Axon.Loop.State{...}, transformed_state}`, so this PR implements said change and updates the tests. This should be the case for other default transforms too? So far I just changed `trainer` but I guess `evaluator` should follow suit. Is that right @seanmor5 ?

Closes #362

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -536,7 +536,7 @@ defmodule Axon.Loop do
     # Build loss now so we can use it as a metric
     loss_fn = build_loss_fn(loss)
     {init_fn, step_fn} = train_step(model, loss_fn, optimizer)
-    output_transform = fn state -> state.step_state[:model_state] end
+    output_transform = fn state -> {state, state.step_state[:model_state]} end
 
     loop =
       step_fn
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -536,7 +536,7 @@ defmodule Axon.Loop do
     # Build loss now so we can use it as a metric
     loss_fn = build_loss_fn(loss)
     {init_fn, step_fn} = train_step(model, loss_fn, optimizer)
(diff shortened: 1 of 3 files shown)


lib/axon/loop.ex
```diff
@@ -536,7 +536,7 @@ defmodule Axon.Loop do
     # Build loss now so we can use it as a metric
     loss_fn = build_loss_fn(loss)
     {init_fn, step_fn} = train_step(model, loss_fn, optimizer)
-    output_transform = fn state -> state.step_state[:model_state] end
+    output_transform = fn state -> {state, state.step_state[:model_state]} end
```
seanmor5, line 539: What I have in mind is keeping this the same `fn state -> state.step_state[:model_state] end`

And rather in `run` returning:

```
{state, output_transform.(state)}
```

Does that make sense?

Ian-GL: Yes, that does make sense. I've made the changes here https://github.com/elixir-nx/axon/pull/382/commits/bd927643760ccf54cec5e177d4b367d8e26aed1b . Should specific tests be added? I've just updated the current ones.

seanmor5:
@Ian-GL I am still considering the implications of this...it may be that we end up essentially copying the model state twice which is inefficient. I will revisit this after I finish a few other things for the Axon release

seanmor5:
@Ian-GL I have thought about this more and unfortunately I don't think this is the best route either. I am going to close this PR for now, but will ping you on #362 when I have a better solution!

PR #386, opened 2022-10-12
Fix ONNX CI

jonatanklosko:
Trying if this works :cat2:

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -110,5 +110,8 @@ jobs:
         with:
           path: tmp/axon
       - name: Run ONNX tests
-        run: |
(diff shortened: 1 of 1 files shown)


seanmor5:
Ah, ONNX is going to be broken until I change the serialization to use the new node format, which I will try to get to this week before we release new versions

jonatanklosko:
@seanmor5 ah I see! Currently the CI fails because `axon` points to a different version of `nx` than `axon_onnx`, so I was trying to work around that, but if it's broken anyway we may as well update the `nx` version when `axon_onnx` is fixed. We can revisit if the version mismatch becomes a problem again.

PR #389, opened 2022-10-14
Add exception when dropout is outside expected range

### LINKED ISSUE #367: Dropout with rate: 1 raises on binary backend

Ian-GL:
The dropout rate is always expected to be in the range of [0, 1). With these changes, now an exception is raised if this happens, instead of letting it crash with an arithmetic error.

Closes #367


seanmor5 (approved):
Great work :)

Ian-GL:
> Great work :)

Thanks, Sean!

PR #401, opened 2022-11-18
Update your_first_axon_model.livemd

joelpaulkoch:
With version 0.3.0 specified this resulted in the following error:

Unchecked dependencies for environment dev:
* nx (https://github.com/elixir-nx/nx.git) the dependency does not match the requirement "~> 0.3.0", got "0.4.0"

So I guess it's a good idea to either remove the version or the github reference (and update the version). Not sure which way you want to follow.

### DIFF
--- a/guides/model_creation/your_first_axon_model.livemd
+++ b/guides/model_creation/your_first_axon_model.livemd
@@ -5,7 +5,7 @@
 ```elixir
 Mix.install([
   {:axon, github: "elixir-nx/axon"},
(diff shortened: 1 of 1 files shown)


seanmor5:
I think going with the GitHub dependency is fine for now :)

seanmor5:
Thanks!

joelpaulkoch:
On the other hand, I just realized that you can select the version in hexdocs. So you'll probably want to fix the versions in the setup accordingly.

PR #406, opened 2022-12-02
Doc latex fixes

chadykamar:
Spotted a couple of issues with some of the LaTeX docs


lib/axon/activations.ex
```diff
@@ -410,7 +410,7 @@ defmodule Axon.Activations do
   @doc ~S"""
   Log-sigmoid activation.
 
-  $$f(x_i) = \log(\sigmoid(x))$$
+  $$f(x_i) = \log(\sigma(x))$$
```
polvalente, line 413: I think the correct fix would be:
```suggestion
  $$f(x_i) = \log(sigmoid(x))$$
```

lib/axon/activations.ex
```diff
@@ -599,7 +599,7 @@ defmodule Axon.Activations do
   @doc ~S"""
   Sigmoid weighted linear unit activation.
 
-  $$f(x_i) = x\sigmoid(x)$$
+  $$f(x_i) = x\sigma(x)$$
```
polvalente, line 602:
```suggestion
  $$f(x_i) = x\ sigmoid(x)$$
```

lib/axon/schedules.ex
```diff
@@ -88,7 +88,7 @@ defmodule Axon.Schedules do
   @doc ~S"""
   Cosine decay schedule.
 
-  $$\gamma(t) = \gamma_0 * (1 - \alpha)*(\frac{1}{2}(1 + \cos{\pi \frac{t}{k}})) + \alpha$$
+  $$\gamma(t) = \gamma_0 * \left(\frac{1}{2}(1 - \alpha)(1 + \cos\pi \frac{t}{k}) + \alpha\right)$$
```
polvalente, line 91: This one seems to correctly reflect the implemented code. @seanmor5 could you check if the code itself isn't what's wrong?

lib/axon/activations.ex
chadykamar: I can change it if you think that's clearer. The sigmoid function can be denoted as σ so I assumed that was the intent here.

lib/axon/schedules.ex
chadykamar: I compared with the paper as well as the Optax implementation and I _think_ the code's correct.

lib/axon/activations.ex
polvalente: This module exposes a `sigmoid` function, so I think `sigmoid` is the clearer alternative with that in mind

polvalente: There's also `Nx.sigmoid` that uses the full name as well

chadykamar: Good point, I amended it.

seanmor5:
Thanks!

PR #412, opened 2022-12-09
Add plot handler

cigrainger:
Adds a function for adding a handler function to update a `VegaLite` plot. Pretty basic -- tracks iterations in the `:x` and a metric in the `:y`.

Before this is ready to go, we need to figure out a better way of handling multiple epochs when we don't know the length of the input data and don't have a `max_iterations` set. Otherwise the `iteration` resets at each new epoch, which causes janky plots as we get duplicate data points per iteration. We could track absolute number of iterations in `Axon.Loop.State`? Maybe I'm missing another way to do this.


lib/axon/loop.ex
```diff
@@ -1304,6 +1304,64 @@ defmodule Axon.Loop do
...
+        |> Vl.encode_field(:x, "step", type: :quantitative)
+        |> Vl.encode_field(:y, "loss", type: :quantitative)
+        |> Kino.VegaLite.new()
+        |> Kino.render()
+
+      model
+      |> Axon.Loop.trainer(loss, optim)
+      |> Axon.Loop.plot(plot)
+  """
+  def plot(loop, plot, metric \\ "loss", event \\ :iteration_completed, filter \\ :always) do
```
josevalim, line 1326:
```suggestion
  def plot(loop, plot, opts \\ []) do
```

and set those defaults with `Keyword.validate!`?

lib/axon/loop.ex
```diff
@@ -1304,6 +1304,64 @@ defmodule Axon.Loop do
     )
   end
 
+  @doc """
```
josevalim, line 1307:
```suggestion
  @compile {:no_warn_undefined, Kino.VegaLite}

  @doc """
```

lib/axon/loop.ex
seanmor5: +1, but I would make `metric` a required argument!

cigrainger: So like this?

```suggestion
  def plot(loop, plot, metric, opts \\ []) do
```

seanmor5: Yep!

seanmor5 (approved):

cigrainger:
@seanmor5 any thoughts on the issue re: iterations? After one iteration the y value resets and we end up with this. Which is... not what we want. 
image

josevalim:
Why does the y value reset? Wouldn't Vega-Lite keep the history as we push?

cigrainger:
Let's say you're streaming data in. Like in the mnist example. If you leave the default -1 max_iterations and run more than epoch, then the iteration value resets to 0 for each epoch. So you send 0..1800 the first epoch, but then those same y values again the next epoch with different x values.

josevalim:
Would a "epoch_iterations" that increments all iterations across epochs address it?

We could also have additional layers to pinpoint whenever a new iteration starts.

josevalim:
Maybe we could even emulate it now by doing `(epoch-1)*max_iterations + iterations`?

cigrainger:
That's a long way of saying the iteration value is the one _within the epoch_ but we want to be sending the one _total across epochs_.

seanmor5:
Oh yeah the reason it resets is we want to keep a history per epoch, but not recompile the step function

We could also keep the history in handle meta and use that here

cigrainger:
@seanmor5 wdyt?

seanmor5:
Thank you!

jonatanklosko:
Awesome!! @seanmor5 @josevalim should we have `kino_axon` perhaps, or are we fine with optional deps and having it here?

PR #417, opened 2022-12-12
Compile batch_fn

josevalim:
This will help us ensure we don't have shapes changing throughout epochs. The current test suite fails exactly because of that.

Since we are now compiling, instead of jitting, it may be that the name `jit_compile?` is no longer relevant. We should discuss if we want to keep the name altogether.


seanmor5:
Merged into #420

PR #418, opened 2022-12-12
Use a more assertive name on plotting

josevalim:
WDYT @cigrainger? This will avoid conflicts in the future if we ever start using another plotting library and removes any ambiguity.


cigrainger:
Sure, totally makes sense to me! Though we could plausibly pattern match on the plot struct. But we don't know what that will be so this makes sense.

josevalim:
That's what I thought but it is only a Kino.JS.Live struct, so we don't know much about it :(

PR #426, opened 2022-12-13
Use a layer state to manage dropout state so it changes between runs

### LINKED ISSUE #423: Dropout keys are too deterministic
Right now we build a bunch of keys up front in layer compilation and pass them to dropout, but this is actually insufficient as the key will be fixed for every computation. That means we'll end up masking the same part of each layer during dropout, which is not at all correct.

Instead, we probably need to treat dropout keys as state and aggregate them during training like we do with some normalization stats

seanmor5:
Resolves #423 

The integration tests are failing because of a mismatched while-body, and I'm not sure where/how that's happening

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -336,8 +336,9 @@ defmodule Axon do
   @doc type: :special
   def param(name, shape, opts \\ [])
       when is_binary(name) and (is_tuple(shape) or is_function(shape)) do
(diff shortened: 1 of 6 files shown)


josevalim:
I pushed another approach to #427. I wonder if the seed should be related to the compile seed/key. And I think it should always be a parameter to it is deterministic.

PR #427, opened 2022-12-13
Provide deterministic builds

josevalim:
* Parameter IDs were removed
* Dropouts are completely removed from the network via a new :mode option
* Freezing traverse the nodes directly without relying on IDs
* Removed almost all usage of backend_copy(Nx.Defn.Expr)
* We use integers as cache keys after the cache is built

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -2879,51 +2879,44 @@ defmodule Axon do
   the update process.
   """
   @doc type: :model
-  def freeze(%Axon{} = model, fun_or_predicate \\ :all) do
-    parameters_per_layer =
-      reduce_nodes(model, [], fn %Axon.Node{parameters: params}, acc ->
-        layer_params =
-          Enum.reduce(params, [], fn param, inner_acc ->
-            [param | inner_acc]
-          end)
+  def freeze(model, fun_or_predicate \\ :all) do
+    freeze(model, fun_or_predicate, true)
+  end
 
-        [layer_params | acc]
-      end)
+  defp freeze(%Axon{output: id, nodes: nodes} = axon, fun_or_predicate, flag) do
+    {nodes, _} = traverse_nodes(id, nodes, [], MapSet.new())
 
-    parameters_to_freeze =
+    nodes =
       case fun_or_predicate do
         :all ->
-          List.flatten(parameters_per_layer)
+          freeze_nodes(nodes, flag)
 
         [{:up, n}] ->
-          parameters_per_layer
-          |> Enum.reverse()
-          |> Enum.take(n)
-          |> List.flatten()
+          {pre, post} = Enum.split(nodes, n)
+          freeze_nodes(pre, flag) ++ post
 
         [{:down, n}] ->
-          parameters_per_layer
-          |> Enum.reverse()
-          |> Enum.drop(n)
-          |> List.flatten()
+          {pre, post} = Enum.split(nodes, -n)
+          pre ++ freeze_nodes(post, flag)
 
         fun ->
-          parameters_per_layer
-          |> List.flatten()
-          |> Enum.filter(fun)
+          Enum.map(nodes, fn %Axon.Node{parameters: params} = axon_node ->
+            %{
+              axon_node
+              | parameters:
+                  Enum.map(params, fn p ->
+                    if fun.(p), do: %{p | frozen: flag}, else: p
+                  end)
+            }
+          end)
       end
 
-    map_nodes(model, fn %Axon.Node{parameters: params} = axon_node ->
-      frozen_params =
-        Enum.map(params, fn %{name: param_name} = v ->
-          if Enum.any?(parameters_to_freeze, fn %{name: name} -> param_name == name end) do
-            %{v | frozen: true}
-          else
-            v
-          end
-        end)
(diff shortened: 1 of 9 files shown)


lib/axon.ex
```diff
@@ -2910,8 +2908,8 @@ defmodule Axon do
 
     map_nodes(model, fn %Axon.Node{parameters: params} = axon_node ->
       frozen_params =
-        Enum.map(params, fn %{id: param_id} = v ->
-          if Enum.any?(parameters_to_freeze, fn %{id: id} -> param_id == id end) do
+        Enum.map(params, fn %{name: param_name} = v ->
+          if Enum.any?(parameters_to_freeze, fn %{name: name} -> param_name == name end) do
```
josevalim, line 2912: @seanmor5 we were using parameters IDs here... is it ok to use names instead?

lib/axon/compiler.ex
```diff
@@ -500,14 +525,26 @@ defmodule Axon.Compiler do
...
-        &to_model_funs(&1, nodes, &2, mode, key)
+        &to_model_funs(&1, nodes, &2, config)
       )
 
     # Names are computed lazily, so compute name from current
     # op and aggregate op_counts.
     name = name_fn.(op_name, op_counts)
     op_counts = Map.update(op_counts, op_name, 1, fn x -> x + 1 end)
 
+    # TODO: Hack for dropout with key, fix with a better implementation
```
josevalim, line 536: Perhaps it should not be based on the name but I would say it should be based on the key. Thoughts?

lib/axon/compiler.ex
```diff
@@ -672,19 +707,9 @@ defmodule Axon.Compiler do
...
-          dropout_key = Nx.Random.fold_in(key, data)
-          opts ++ [key: dropout_key]
-        else
-          opts
-        end
-
       # Compute arguments to be forwarded and ensure `:mode` is included
       # for inference/training behavior dependent functions
-      args = Enum.reverse(tensor_inputs) ++ [Keyword.put(opts, :mode, mode)]
+      args = Enum.reverse(tensor_inputs, [Keyword.put(opts, :mode, mode)])
```
josevalim, line 712: `Enum.reverse(list, tail)` is an efficient version of `Enum.reverse(list) ++ tail`.

lib/axon/layers.ex
```diff
@@ -1627,15 +1627,7 @@ defmodule Axon.Layers do
...
-    out = Nx.select(mask, input / keep_prob, Nx.tensor(0, type: Nx.type(input)))
-
-    transform({input, out, opts[:mode]}, fn
-      {input, _, :inference} ->
-        input
-
-      {_, out, :train} ->
-        out
-    end)
+    Nx.select(mask, input / keep_prob, Nx.tensor(0, type: Nx.type(input)))
```
josevalim, line 1630: I removed the mode check from dropout because it is no longer relevant.

lib/axon/losses.ex
```diff
@@ -289,6 +282,7 @@ defmodule Axon.Losses do
       >
 
   """
+  # TODO: Weights must be a third argument instead of an option
```
josevalim, line 285: It doesn't affect this PR for now but we want to fix this.

test/axon/compiler_test.exs
```diff
@@ -114,7 +114,7 @@ defmodule CompilerTest do
       x2 = Axon.dense(input, 64)
       model = Axon.add(x1, x2)
 
-      {init_fn, _predict_fn} = Axon.build(model)
+      {init_fn, _predict_fn} = Axon.build(model, debug: true)
```
josevalim, line 117: We only get fn stacktraces with `debug: true` now.

josevalim:
This is good to go!

lib/axon/losses.ex
josevalim: I will fix this in another PR.

lib/axon/losses.ex
```diff
@@ -135,19 +135,12 @@ defmodule Axon.Losses do
...
-
-        {y_true, pos, neg} ->
-          Nx.take(Nx.tensor([neg, pos], backend: Nx.Defn.Expr), y_true)
-      end)
+      case {opts[:positive_weight], opts[:negative_weight]} do
+        {nil, nil} -> nil
+        {pos, nil} -> Nx.take(Nx.tensor([1.0, pos]), y_true)
+        {nil, neg} -> Nx.take(Nx.tensor([neg, 1.0]), y_true)
+        {pos, neg} -> Nx.take(Nx.tensor([neg, pos]), y_true)
+      end
```
josevalim, line 143: I will do this in another PR:

```suggestion
      transform({y_true, opts[:positive_weight], opts[:negative_weight]}, fn
        {_, nil, nil} ->
          nil

        {y_true, pos, nil} ->
          Nx.take(Nx.tensor([1.0, pos], backend: Nx.Defn.Expr), y_true)

        {y_true, nil, neg} ->
          Nx.take(Nx.tensor([neg, 1.0], backend: Nx.Defn.Expr), y_true)

        {y_true, pos, neg} ->
          Nx.take(Nx.tensor([neg, pos], backend: Nx.Defn.Expr), y_true)
      end)
```

lib/axon/losses.ex
josevalim, line 285:
```suggestion
```

PR #439, opened 2023-01-04
Fix Fashion MNIST autoencoder example

nskins:
The Fashion MNIST autoencoder example throws the same error as the examples in #437. This PR fixes the autoencoder by updating the dependencies and fixing breaking changes.

Summary of changes:
- Update dependencies to latest versions
- Add `:batch` placeholder for dynamic batch size
- Update deprecated `Nx` function `to_batched_list/3` to `to_batched/3`
- Use `Enum.fetch!/2` since `hd/1` does not work on a stream
- Remove unused alias for `Axon.Loop.State`
- Fix spelling of MNIST in module name

### DIFF
--- a/examples/generative/fashionmnist_autoencoder.exs
+++ b/examples/generative/fashionmnist_autoencoder.exs
@@ -57,7 +57,7 @@ defmodule FashionMNIST do
 
     sample_image =
       train_images
-      |> hd()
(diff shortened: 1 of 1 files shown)


seanmor5:
Thanks! @nskins

PR #441, opened 2023-01-06
Fix text generator example

nskins:
This PR resolves an incompatibility conflict amongst the dependencies in the text generator example. I think we should prevent this in the future by setting each dependency to a specific version. This PR also fixes breaking changes resulting from the updated dependencies.

Summary of changes:
- Update dependencies to latest versions
- Update output sequence to first element of tuple returned by `Axon.lstm/2`
- Update deprecated `Nx` function `to_batched_list/3` to `to_batched/3`


seanmor5:
Thank you!

PR #445, opened 2023-01-13
Include the folder for the guide links

matthusby:
The guides readme links were broken.  This PR updates the links to the livemd version and include the correct folder its contained in.


josevalim (approved):
+1. This will also work when published!

PR #450, opened 2023-01-16
Fix issues with mismatched templates

### LINKED ISSUE #443: Dropout keys cause compilation mismatches
Dropout keys are somehow getting converted from u32 to f32 in a template argument during training loop, which makes it impossible to run a training loop with strict compilation enabled (to avoid recompilations)

seanmor5:
Resolves #443

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -351,7 +351,7 @@ defmodule Axon.Loop do
           loss: Nx.tensor(0.0),
           gradient_step: Nx.tensor(0),
           model_state: model_state,
-          gradient_state: zeros_like(model_state),
+          gradient_state: zeros_like(model_state, type: :f32),
           optimizer_state: optimizer_state,
           loss_scale_state: loss_scale_state
         }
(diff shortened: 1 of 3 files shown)


lib/axon/loop.ex
```diff
@@ -458,30 +458,26 @@ defmodule Axon.Loop do
...
+          new_model_state = Axon.Updates.apply_updates(model_state, updates, new_state)
+
+          {gradients, new_model_state, new_state, new_optimizer_state, new_gradient_state, 0,
+           Nx.tensor(0)}
+        else
+          acc_gradients = deep_merge(gradient_state, gradients, fn x, y -> x + y end)
+
+          {gradients, model_state, new_state, optimizer_state, acc_gradients, gradient_step + 1,
+           Nx.tensor(0)}
+        end
```
josevalim, line 479: Wait, both branches return `Nx.tensor(0)`. Doesn't it mean you ever only do a single pass anyway?

seanmor5: Yes, I'm pretty sure this is how I got around larger graphs with if, iirc this reduces memory usage slightly

PR #454, opened 2023-01-17
Fix deprecation warnings with new Nx

seanmor5:
Cleaning up final blockers before next release

### DIFF
--- a/lib/axon/schedules.ex
+++ b/lib/axon/schedules.ex
@@ -22,7 +22,6 @@ defmodule Axon.Schedules do
   """
 
   import Nx.Defn
-  import Axon.Shared
 
   @doc """
   Linear decay schedule.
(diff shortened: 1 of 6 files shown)


josevalim:
The warnings for Nx.tensor should be fixed by passing those as arguments instead of `opts`. That's easy for everything except for layers. I assume we don't want to make the `opts[:rate]` an argument, right? So the current fix is fine. I will push a commit for the other ones. :)

lib/axon/schedules.ex
```diff
@@ -166,7 +165,7 @@ defmodule Axon.Schedules do
 
   defnp apply_constant(_step, opts \\ []) do
     opts = keyword!(opts, init_value: 0.01)
-    Nx.tensor(opts[:init_value])
+    opts[:init_value]
```
josevalim, line 168: I wonder if we should implement constant as literally:

```
def constant(init_value, _opts \\ []) do
  Nx.tensor(init_value)
end
```

?

josevalim (approved):
Besides the two minor comments, it looks good to me. I pushed a commit to address the initializers bit. :)

PR #461, opened 2023-01-21
Start to add stream to loop

seanmor5:


lib/axon/loop.ex
```diff
@@ -1621,134 +1640,212 @@ defmodule Axon.Loop do
...
 
-    {_, state} = fire_event(status, handler_fns, state, debug?)
-    state = %State{state | metrics: final_metrics}
+    {_, state} =
+      case state do
+        %{status: :completed} = state -> fire_event(:completed, handler_fns, state, debug?)
+        %{status: :halted} = state -> fire_event(:halted, handler_fns, state, debug?)
+      end
 
     output_transform.(state)
```
josevalim, line 1809: Good opportunity to kill `output_transform`.

lib/axon/loop.ex
```diff
@@ -1621,134 +1640,212 @@ defmodule Axon.Loop do
...
-      end
+      end)
 
-    {_, state} = fire_event(status, handler_fns, state, debug?)
-    state = %State{state | metrics: final_metrics}
+    {_, state} =
+      case state do
+        %{status: :completed} = state -> fire_event(:completed, handler_fns, state, debug?)
+        %{status: :halted} = state -> fire_event(:halted, handler_fns, state, debug?)
+      end
```
josevalim, line 1807: We should convert this to `on_done` cause this state cannot be captured from a stream.

PR #467, opened 2023-01-24
Remove deprecated transform

### LINKED ISSUE #463: Remove use of deprecated transforms

seanmor5:
Resolves #463


lib/axon/initializers.ex
```diff
@@ -722,6 +696,33 @@ defmodule Axon.Initializers do
...
+        :uniform ->
+          Nx.Random.uniform_split(key, 0.0, 1.0, shape: flat_shape, type: type)
+
+        :normal ->
+          Nx.Random.normal_split(key, 0.0, 1.0, shape: flat_shape, type: type)
+
+        dist ->
+          raise ArgumentError,
+                "invalid distribution #{inspect(dist)} passed to orthogonal/1"
+      end
```
josevalim, line 721: FWIW, I think this could remain as part of `defn`. Not sure.

lib/axon/losses.ex
```diff
@@ -1134,6 +1123,14 @@ defmodule Axon.Losses do
     t0_prob
   end
 
+  deftransformp ctc_reduction(loss, l_true, reduction) do
+    case {reduction, loss} do
+      {:mean, loss} -> Nx.divide(loss, l_true) |> Nx.mean()
+      {:sum, loss} -> Nx.sum(loss)
+      {:none, loss} -> loss
+    end
```
josevalim, line 1131: This might just work in `defn` too:

    case reduction do
      :mean -> Nx.divide(loss, l_true) |> Nx.mean()
      :sum -> Nx.sum(loss)
      :none -> loss
    end

lib/axon/metrics.ex
```diff
@@ -449,7 +442,11 @@ defmodule Axon.Metrics do
     end
   end
 
-  defnp(top_k_index_transform(y_true), do: Nx.argmax(y_true, axis: -1, keep_axis: true))
+  deftransformp sparse_transform(y_true, sparse) do
+    if sparse, do: y_true, else: top_k_index_transform(y_true)
```
josevalim, line 446: This should also work in `defn`.

lib/axon/losses.ex
josevalim: To be clear, `case atom do` should work on `defn`. :)

lib/axon/initializers.ex
josevalim: the `case distribution do`, to be more precise.

PR #468, opened 2023-01-25
docs: improve documentation on halt_epoch

polvalente:


lib/axon/loop.ex
```diff
@@ -161,6 +161,10 @@ defmodule Axon.Loop do
       :halt_epoch # Halt the epoch, continue looping
       :halt_loop  # Halt looping
 
+  It is important to notice that `:halt_epoch` does the same
+  as `:continue` if not accompanied by an `:epoch_halted` handler
+  that also emits a `:halt_epoch` instruction.
```
seanmor5, line 166: Wait, I don't think this is intended, can you expand?

seanmor5: `:halt_epoch` should do exactly that where it halts the epoch, fires `:epoch_halted`, and then short circuts, so there may be a bug

polvalente: If :halt_epoch is emmited, an :epoch_halted is triggered, but if you don't "catch" it, it's as if you didn't halt. It's due to how the code is structured, that when a :halt_epoch is emitted a raw and stateless :epoch_halted is triggered and its result is passed forward

seanmor5: Oh yeah then that is a bug and should be changed, it will be addressed in #461

polvalente: Right. Close this if you feel like this shouldn't be documented/tested in the meantime.

In that case, I'll just restructure the test I edited to reflect the timing assertions that I added.

The current test only asserts on amounts, and not on the order of events.

PR #470, opened 2023-01-30
Deprecate Axon.Loop.handle/4

seanmor5:
Goes along with streaming. The differentiator is to possibly implement some mechanism for 2 way communication with a loop running async


josevalim:
@seanmor5 is this just a name change? We should be able to keep the cont/halt semantics even if running async. The issue is that by the time the halt is received, you will have to process at least one other batch/epoch.

seanmor5:
Yeah this is just a name change, the main thing is that `handle_event` will be something explicitly fired in a loop whereas there may be an alternative `handle_message` or something to handle an external message sent from the caller

PR #473, opened 2023-02-05
docs: update `channels` default value mentioned in the docs

nickgnd:
Hey 👋 
I'm opening this small PR to fix a discrepancy in the docs for the option `channels` used in the Conv layers and others.

The default value for the `:channels` option is `:last` based on what I can see in the code (e.g. [here](https://github.com/elixir-nx/axon/blob/7c21ae75d34e68c946a8ff24a46f194f9d17003a/lib/axon/layers.ex#L356)).

Note that in the `Axon` module docs, the default value for the option is correct, only in the `Axon.Layers` it was outdated/incorrect (`:first`).

Cheers ✌️

### DIFF
--- a/lib/axon/layers.ex
+++ b/lib/axon/layers.ex
@@ -241,7 +241,7 @@ defmodule Axon.Layers do
       Defaults to `1` or no dilation.
 
     * `:channels ` - channel configuration. One of `:first` or `:last`.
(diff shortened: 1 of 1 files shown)


seanmor5 (approved):
Thank you!

PR #484, opened 2023-04-10
Add huber loss

seanmor5:
@polvalente

### DIFF
--- a/lib/axon/losses.ex
+++ b/lib/axon/losses.ex
@@ -972,7 +972,8 @@ defmodule Axon.Losses do
     * `:reduction` - reduction mode. One of `:mean`, `:sum`, or `:none`.
       Defaults to `:none`.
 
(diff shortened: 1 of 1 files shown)


lib/axon/losses.ex
```diff
@@ -959,6 +959,55 @@ defmodule Axon.Losses do
...
+
+    * `y_true` - $(d_0, d_1, ..., d_n)$
+    * `y_pred` - $(d_0, d_1, ..., d_n)$
+
+  ## Options
+
+    * `:reduction` - reduction mode. One of `:mean`, `:sum`, or `:none`.
+      Defaults to `:none`.
+
+    * `:delta` - clip for values larger than `delta`
```
polvalente, line 975: This isn't what delta does. TF has a nice formula to show what delta does

polvalente (approved):

PR #492, opened 2023-04-26
fix: decay defaults for optimizers

polvalente:
Fixes defaults for :decay option in some optimizers.
RMSProp wasn't using the option, but I'm not sure if momentum is being used correctly there

### DIFF
--- a/lib/axon/updates.ex
+++ b/lib/axon/updates.ex
@@ -239,7 +239,7 @@ defmodule Axon.Updates do
 
   ## Options
 
-      * `:decay` - EMA decay rate. Defaults to `0.9`.
+      * `:decay` - EMA decay rate. Defaults to `0.95`.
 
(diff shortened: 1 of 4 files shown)


lib/axon/updates.ex
```diff
@@ -771,7 +771,7 @@ defmodule Axon.Updates do
   Adds random Gaussian noise to the input.
```
polvalente, line 771: The `trace` function in this file uses a decay-like algorithm with `:decay` defaulting to `0.9`. Should we use this default throughout?

polvalente:
@seanmor5 Optax actually uses adamw decay ~ 0 by default, instead of ~ 1. It's best if you validate these defaults before reviewing the PR. Maybe the tests are checking agains unusual values.

https://optax.readthedocs.io/en/latest/api.html#optax.adamw

seanmor5:
@polvalente if you add a test without the defaults do they pass?

PR #498, opened 2023-06-02
Fix precision issue

### LINKED ISSUE #496: How to verify mixed precision is working for language models
Originally I was [reading](https://lightning.ai/pages/community/tutorial/accelerating-large-language-models-with-mixed-precision-techniques/) about how people do this type of mixed precision work in python today and found it was a configuration like you see here
    
Screenshot 2023-05-26 at 10 19 28 AM

Wanting to accomplish this in Axon I started by looking at [this pull request ](https://github.com/elixir-nx/axon/pull/74/files)about mixed precision to use a `bf: 16` policy when fine tuning BERT but it doesn't appear to be working and I'm looking for clues to prove/disprove it.

I noticed in the [training test code](https://github.com/elixir-nx/axon/blob/b7b19c61ca0f242a8be4f02d87f0657146609f29/test/mixed_precision_test.exs#L62) for `correctly maintains parameter type after train step` the assertions verify the params after training so I went ahead and did the same. But in my example it shows I'm still using `f: 32` for some odd reason.

```
%{
  "bias" => #Nx.Tensor<
    f32[768]
    EXLA.Backend<cuda:0, 0.198458620.3614310435.139911>
    [1.1483082926133648e-4, 1.5510240336880088e-4, 8.907304618333001e-7, 4.598435771185905e-5, 7.280074351001531e-5, 3.678119392134249e-4, -1.5480619913432747e-4, -6.495109118986875e-5,-0.015686072409152985, ...]
  >,
  "kernel" => #Nx.Tensor<
    f32[768][768]
    EXLA.Backend<cuda:0, 0.198458620.3614310435.139912>
    [
      [0.09109397977590561, -0.018297171220183372, 0.01780594326555729, 0.03939569741487503, 0.12485719472169876, 7.235617376863956e-4, 0.01607041247189045, 0.07742417603731155, ...],
      ...
    ]
  >
}
```

I also found no reduction in memory at runtime (using the configuration that avoids using all GPU VRAM until it's truly needed) which makes me suspicious that this policy is truly applied. I assume it has more to do with the model and my fine tuning than the mixed precision code implemented in that PR so here is my approach.

Here is the full source for my fine tuning example with this mixed precision policy applied

```elixir
    ## fine tune bert
    %{model: model, params: params} = model

    [{input, _}] = Enum.take(train_data, 1)
    Axon.get_output_shape(model, input)

    logits_model = Axon.nx(model, & &1.logits)

    loss =
      &Axon.Losses.categorical_cross_entropy(&1, &2,
        reduction: :mean,
        from_logits: true,
        sparse: true
      )

    optimizer = Axon.Optimizers.adam(2.0e-5)
    accuracy = &Axon.Metrics.accuracy(&1, &2, from_logits: true, sparse: true)

    policy = Axon.MixedPrecision.create_policy(
      params: {:bf, 16},
      compute: {:bf, 16},
      output: {:bf, 16}
    )

    trained_model_state =
      logits_model
      |> Axon.MixedPrecision.apply_policy(policy, except: [:batch_norm])
      |> Axon.Loop.trainer(loss, optimizer, log: 1)
      |> Axon.Loop.metric(accuracy, "accuracy")
      |> Axon.Loop.run(train_data, params, epochs: 1, compiler: EXLA, strict?: false)

    # this is how I pulled the params to see if they are f32/bf16
    trained_model_state
    |> Map.keys()
    |> Enum.take(1)
    |> Enum.each(fn st ->
      trained_model_state[st] |> IO.inspect(label: "compare this...")
    end)
```

seanmor5:
Resolves #496 

When a user has a pretrained model with some initialized state, we do not cast that state to the model's desired policy if they try to use mixed precision. This fixes that.

It kind of makes me think though that maybe we should rethink mixed precision a bit? I'm not exactly sure how, but doing this implicitly feels weird to me. Maybe the alternative is to recommend the parameters in this case be cast to the desired type explicitly.

### DIFF
--- a/lib/axon/compiler.ex
+++ b/lib/axon/compiler.ex
@@ -203,13 +203,26 @@ defmodule Axon.Compiler do
           %{params | key => merge_params!(nested, value)}
 
         %{^key => template} ->
-          %{params | key => Nx.as_type(value, Nx.type(template))}
+          %{params | key => merge_type(key, template, value)}
 
         _ ->
           Logger.warning("found unexpected key in the initial parameters map: #{inspect(key)}")
           params
       end
     end)
+  end
+
+  defp merge_type(key, template, value) do
+    if Nx.type(template) != Nx.type(value) do
+      Logger.warning(
+        "initial type for parameter #{key} does not match policy," <>
+          " consider using Axon.MixedPrecision.cast before passing" <>
+          " initial state to model initialization function to avoid" <>
+          " type casts"
+      )
+    end
+
+    Nx.as_type(value, Nx.type(template))
   end
 
   def compile(graph, _opts) do
--- a/lib/axon/compiler.ex
+++ b/lib/axon/compiler.ex
@@ -202,8 +202,8 @@ defmodule Axon.Compiler do
         %{^key => %{} = nested} when not is_struct(nested) ->
           %{params | key => merge_params!(nested, value)}
 
-        %{^key => _} ->
-          %{params | key => value}
+        %{^key => template} ->
+          %{params | key => Nx.as_type(value, Nx.type(template))}
 
         _ ->
           Logger.warning("found unexpected key in the initial parameters map: #{inspect(key)}")
--- a/lib/axon/mixed_precision.ex
+++ b/lib/axon/mixed_precision.ex
@@ -41,6 +41,7 @@ defmodule Axon.MixedPrecision do
   """
 
   alias Axon.MixedPrecision.Policy
+  import Axon.Shared
 
   @doc """
   Creates a mixed precision policy with the given options.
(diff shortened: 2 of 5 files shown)


josevalim:
@seanmor5 maybe an explicit `Axon.cast(model, params)` function? Although I think implicit casting is fine cause we want to move the casting itself as part of the model? What do other libraries/frameworks do?

seanmor5:
I actually think a good route is:

- Add Axon.MixedPrecision.cast(policy, container_or_tensor, type) which casts the container according to the policy
- Warn in `merge_params!` when initialized type differs from the input type and suggest explicitly casting to avoid issues in the future

PR #501, opened 2023-06-07
refactor: deprecate optimization modules in favor of Optimus

polvalente:


lib/axon/optimizers.ex
```diff
@@ -1,58 +1,5 @@
...
-
-  For a simpler approach, you can also use optimizers with the training API:
-
-        model
-        |> Axon.Loop.trainer(:categorical_cross_entropy, Axon.Optimizers.adam(0.005))
-        |> Axon.Loop.run(data, epochs: 10, compiler: EXLA)
-
-  """
+  @moduledoc false
   alias Axon.Updates
```
josevalim, line 3:
```suggestion
  alias Optimus.Updates
```

?

josevalim (approved):

lib/axon/optimizers.ex
polvalente: I don't think we should do this. I believe `axon/{optimizers,updates,schedules}` should be left in this kind of "old version" kind of state until they are fully removed.

josevalim: I don't think we have an option, otherwise Axon will emit one bazilion warnings when compiling. Given the implementation is the same, I think we should do this change.

polvalente: Should we just fully delegate the whole modules then? Like with macro witchcraft

josevalim: We can do straight-forward delegation, I am not sure if it is worth macro-ing it, but I am also fine with leaving it as is. I would basically do the minimum amount of changes so this code is deprecated and compiles without warnings (as it is effectively dead code).

PR #503, opened 2023-06-09
chore: use :polaris from Hex

polvalente:


polvalente:
back to draft because I forgot that we need to change the learning rates into an option now :)

seanmor5 (approved):

PR #504, opened 2023-06-21
fix loss_scale atom typo in Axon.Loop.trainer

wtedw:
The trainer function is currently looking for :loss_cale in opts. This should pick out the right options and pass them to train_step


seanmor5:
Good catch!

PR #505, opened 2023-06-26
Fixes a few warnings in the example scripts

grzuy:
Hi, 

Minor changes to the examples.

Thanks for your hard work on this library!

### DIFF
--- a/examples/vision/cifar10.exs
+++ b/examples/vision/cifar10.exs
@@ -24,7 +24,6 @@ defmodule Cifar do
     |> Nx.new_axis(-1)
     |> Nx.equal(Nx.tensor(Enum.to_list(0..9)))
     |> Nx.to_batched(32)
-    |> Enum.to_list()
     |> Enum.split(1500)
   end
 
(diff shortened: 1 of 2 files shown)


examples/vision/cifar10.exs
```diff
@@ -22,7 +23,8 @@ defmodule Cifar do
     |> Nx.from_binary(type)
     |> Nx.new_axis(-1)
     |> Nx.equal(Nx.tensor(Enum.to_list(0..9)))
-    |> Nx.to_batched_list(32)
+    |> Nx.to_batched(32)
+    |> Enum.to_list()
```
seanmor5, line 27: I think `Enum.split` will just do this implicitly
```suggestion
```

examples/vision/cifar10.exs
```diff
@@ -13,7 +13,8 @@ defmodule Cifar do
     |> Nx.from_binary(type)
     |> Nx.reshape({elem(shape, 0), 32, 32, 3})
     |> Nx.divide(255.0)
-    |> Nx.to_batched_list(32)
+    |> Nx.to_batched(32)
+    |> Enum.to_list()
```
seanmor5, line 17: Same as below:

```suggestion
```

seanmor5 (approved):

examples/vision/cifar10.exs
grzuy: Good call, thanks!!

PR #506, opened 2023-06-27
Fixes horses_or_humans.exs example

grzuy:
- channels transpose no longer needed (it seems?)
- Polaris references in example needs polaris dependency until next axon is released
- Fix erroneous png parsing
- Updates dataset source URL

### DIFF
--- a/examples/vision/horses_or_humans.exs
+++ b/examples/vision/horses_or_humans.exs
@@ -10,7 +10,8 @@ defmodule HorsesOrHumans do
   alias Axon.Loop.State
   import Nx.Defn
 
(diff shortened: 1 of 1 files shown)


examples/vision/horses_or_humans.exs
```diff
@@ -1,6 +1,7 @@
 Mix.install([
   {:stb_image, "~> 0.5.2"},
   {:axon, "~> 0.5"},
+  {:polaris, "~> 0.1.0"},
```
polvalente, line 4:
```suggestion
  {:polaris, "~> 0.1"},
```

seanmor5:
Thanks!

PR #508, opened 2023-06-28
Fixes credit_card_fraud.exs example

grzuy:
Seems the example was written for `explorer 0.2` and never adapted to support latest explorer.

For now quick fix to go back to compatible explorer version.
Haven't had time to see which/how many updates necessary to make it work with explorer 0.5.

Thanks!

### DIFF
--- a/examples/structured/credit_card_fraud.exs
+++ b/examples/structured/credit_card_fraud.exs
@@ -126,7 +126,7 @@ defmodule CreditCardFraud do
     model
     |> Axon.Loop.evaluator()
     |> metrics()
-    |> Axon.Loop.handle(:epoch_completed, &summarize/1)
+    |> Axon.Loop.handle_event(:epoch_completed, &summarize/1)
     |> Axon.Loop.run(test_data, model_state, compiler: EXLA)
   end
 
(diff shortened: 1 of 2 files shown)


seanmor5:
I think we should make an effort to make this compatible with Explorer 0.5 instead of keeping it at Explorer 0.2, especially given how much has been changed/added

grzuy:
OK.

I think I found a way to normalize data frame in a more concise and efficient way using latest `explorer` features, but it depends on unreleased `abs` function.

So this PR, as of current implementation, is blocked on having an `explorer` release first.

examples/structured/credit_card_fraud.exs
```diff
@@ -1,12 +1,14 @@
 Mix.install([
   {:axon, "~> 0.5"},
+  {:polaris, "~> 0.1"},
   {:exla, "~> 0.5"},
   {:nx, "~> 0.5"},
-  {:explorer, "~> 0.5"}
+  {:explorer, path: "../explorer"} # 0.5.8-dev
```
grzuy, line 6: This won't work.
Pending update to future explorer release.

grzuy: Updated now to point to latest explorer release.
https://github.com/elixir-nx/axon/pull/508/commits/38135fa45bef6c2d30bb477d4fa2c91ee1f97d88

PR #512, opened 2023-07-17
refactor(examples): attempt to improve clarity of where some hard-coded numbers coming from or mean

grzuy:
Thought something like this may help improve understanding of these model examples for first-time readers.

From my own experience when first reading these a while ago, the most challenging thing at first was understanding what hard-coded values meant or where they were coming from, and how and if they where related with each other.


grzuy:
Test failure seems to be also present in last `main` commit.

examples/generative/fashionmnist_autoencoder.exs
```diff
@@ -31,9 +38,9 @@ defmodule FashionMNIST do
   defp transform_images({bin, type, shape}) do
     bin
     |> Nx.from_binary(type)
-    |> Nx.reshape({elem(shape, 0), 1, 28, 28})
-    |> Nx.divide(255.0)
-    |> Nx.to_batched(32)
+    |> Nx.reshape({elem(shape, 0), @image_channels, @image_side_pixels, @image_side_pixels})
```
polvalente, line 41:
```suggestion
    |> Nx.reshape({elem(shape, 0), 1, 28, 28}, names: [:batch, :channels, :width, :height])
```

not an _actual_ suggestion/review, just so you know of this other way of imposing semantics to shapes

examples/basics/multi_input_example.exs
```diff
@@ -18,8 +20,8 @@ defmodule XOR do
   end
 
   defp batch do
-    x1 = Nx.tensor(for _ <- 1..32, do: [Enum.random(0..1)])
-    x2 = Nx.tensor(for _ <- 1..32, do: [Enum.random(0..1)])
+    x1 = Nx.tensor(for _ <- 1..@batch_size, do: [Enum.random(0..1)])
+    x2 = Nx.tensor(for _ <- 1..@batch_size, do: [Enum.random(0..1)])
```
msluszniak, line 24:
```suggestion
    {x1, new_key} = Nx.Random.randint(key(), 0, 1 + 1, shape: {@batch_size})
    {x2, _new_key} = Nx.Random.randint(new_key, 0, 1 + 1, shape: {@batch_size})
```

examples/basics/multi_input_example.exs
```diff
@@ -7,6 +7,8 @@ Mix.install([
 defmodule XOR do
   require Axon
 
+  @batch_size 32
+
```
msluszniak, line 11:
```suggestion
  @batch_size 32
  
  defp key do
    Nx.Random.key(42)
  end
```

examples/vision/horses_or_humans.exs
```diff
@@ -29,7 +33,7 @@ defmodule HorsesOrHumans do
 
   defnp augment(inp) do
     # Normalize
-    inp = inp / 255.0
+    inp = inp / Nx.Constants.max(@input_type)
```
msluszniak, line 36: For me, this change makes it more confusing. The fact that the max value of the `u8` type is indeed 255 and the max value of `inp` is also 255 is rather an incidence. In my opinion, the comment that the maximum value is 255.0 so we normalize to [0,1] interval by dividing by 255.0 would be better.

examples/basics/multi_input_example.exs
grzuy: Thanks for the suggestion.

Just realized after applying it and testing it locally that as is it generates always the same batches because of the batch always starting with the same key for every batch.

In order to make it work really randomly we need to keep the returned new keys feed back into the batch generation while streaming data I guess.

msluszniak: Yes, you're right

examples/vision/horses_or_humans.exs
grzuy: Right... good call!
Will update.

grzuy: I figured maybe something like this https://github.com/elixir-nx/axon/pull/512/commits/13f6662c2bf3d2b1eccef989de9f381ac2d391b6 could help in making it clear this is the maximum possible integer value of each image pixel channel.

msluszniak: Yeap, this looks good :)

seanmor5:
Thanks!

PR #513, opened 2023-07-20
refactor(examples): prefer Nx.Random over Enum.random

grzuy:
Attempt to address suggestion https://github.com/elixir-nx/axon/pull/512#discussion_r1265913601.

### DIFF
--- a/examples/basics/multi_input_example.exs
+++ b/examples/basics/multi_input_example.exs
@@ -22,8 +22,8 @@ defmodule XOR do
   end
 
   defp batch do
-    {x1, new_key} = Nx.Random.randint(key(), 0, 1 + 1, shape: {32})
-    {x2, _new_key} = Nx.Random.randint(new_key, 0, 1 + 1, shape: {32})
+    {x1, new_key} = Nx.Random.randint(key(), 0, 1 + 1, shape: {32, 1})
+    {x2, _new_key} = Nx.Random.randint(new_key, 0, 1 + 1, shape: {32, 1})
     y = Nx.logical_xor(x1, x2)
     {%{"x1" => x1, "x2" => x2}, y}
   end
--- a/examples/basics/multi_input_example.exs
+++ b/examples/basics/multi_input_example.exs
@@ -18,8 +18,8 @@ defmodule XOR do
   end
 
   defp batch(key) do
-    {x1, next_key} = Nx.Random.randint(key, 0, 1 + 1, shape: {32, 1})
-    {x2, next_key} = Nx.Random.randint(next_key, 0, 1 + 1, shape: {32, 1})
+    {x1, random_key} = Nx.Random.randint(random_key, 0, 2, shape: {32, 1})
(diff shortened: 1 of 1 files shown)


examples/basics/multi_input_example.exs
```diff
@@ -17,11 +17,12 @@ defmodule XOR do
     |> Axon.dense(1, activation: :sigmoid)
   end
 
-  defp batch do
-    x1 = Nx.tensor(for _ <- 1..32, do: [Enum.random(0..1)])
-    x2 = Nx.tensor(for _ <- 1..32, do: [Enum.random(0..1)])
+  defp batch(key) do
+    {x1, next_key} = Nx.Random.randint(key, 0, 1 + 1, shape: {32, 1})
+    {x2, next_key} = Nx.Random.randint(next_key, 0, 1 + 1, shape: {32, 1})
```
polvalente, line 22: _Normally_, I prefer to avoid variable name rebinding in Elixir. However, I like keeping the same name for the random_key variable to ensure that I'm always using the latest state, so I think we should use this in the examples.

```suggestion
    {x1, random_key} = Nx.Random.randint(random_key, 0, 1 + 1, shape: {32, 1})
    {x2, random_key} = Nx.Random.randint(random_key, 0, 1 + 1, shape: {32, 1})
```

grzuy:
Running this locally, before and after, noticed it takes way more time now...

examples/basics/multi_input_example.exs
msluszniak: I originally write 1+1 because it gives you sugestion that the interval in randint is rhs exclusive, simple 2 does not gives you this sugestion. This is a special case where it is not so clear but I believe that for interval [0, 100] if you write randint(0, 100+1) it is much clearer.

polvalente:
@grzuy probably because you're running using Nx.BinaryBackend. Nx.Random has lots of functions chained together, which end up being slow in the default backend. So it's expected to be slower in that sense, but it's also the correct way of doing things if working with random state inside defn.

In this case however, I'm ok with not using Nx.Random because the random calls are just being used to generate random data.

examples/basics/multi_input_example.exs
polvalente: Even if it were 100+1 I would still be caught up in that for quite some time. It's probably better to add a comment prefacing with exactly the reason why we have 2 (or 101 in the other case).

We can also make the docs for Nx.Random more clear because they're just using math interval notation instead of explaining in English

polvalente: https://github.com/elixir-nx/nx/pull/1263

grzuy:
> @grzuy probably because you're running using Nx.BinaryBackend. Nx.Random has lots of functions chained together, which end up being slow in the default backend. So it's expected to be slower in that sense, but it's also the correct way of doing things if working with random state inside defn.

Gotcha, makes sense :+1:

> 
> In this case however, I'm ok with not using Nx.Random because the random calls are just being used to generate random data.

Right, just realized batch is not defn.

Closing, thanks.

PR #515, opened 2023-07-26
ci: tests against 1.15/OTP26 pair

grzuy:


grzuy:
Test failing in OTP26 job seems to be legit failure in `test/axon/compiler_test.exs:5388`, I can reproduce also running it locally.

grzuy:
> Test failure in OTP26 job seems to be legit failure in `test/axon/compiler_test.exs:5388`, I can reproduce also running it locally.

And the EXLA job failure is consistent to the one already present in `main`.

grzuy:
> Test failure in OTP26 job seems to be legit failure in `test/axon/compiler_test.exs:5388`, I can reproduce also running it locally.

Seems that undeterministic map order from OTP26 is affecting this test case

https://github.com/elixir-nx/axon/blob/713e44051854871e994260e14498cff59fd65758/test/axon/compiler_test.exs#L5384-L5391

which relies on `Map.from_struct` keeping order of struct fields...

https://github.com/elixir-nx/axon/blob/713e44051854871e994260e14498cff59fd65758/lib/axon/layers.ex#L2526-L2529

Still not sure what should be the expected behavior of `Axon.stack_columns` for OPT26+... :thinking:

grzuy:
> > Test failure in OTP26 job seems to be legit failure in `test/axon/compiler_test.exs:5388`, I can reproduce also running it locally.
> 
> Seems that undeterministic map key order in OTP26 is affecting this test case
> 
> https://github.com/elixir-nx/axon/blob/713e44051854871e994260e14498cff59fd65758/test/axon/compiler_test.exs#L5384-L5391
> 
> which relies on `Map.from_struct` keeping order of struct fields...
> 
> https://github.com/elixir-nx/axon/blob/713e44051854871e994260e14498cff59fd65758/lib/axon/layers.ex#L2526-L2529
> 
> Still not sure what should be the expected behavior of `Axon.stack_columns` for OPT26+... 🤔

Hacky solution to keep `Axon.stack_columns` relying on the order of the container struct `defstruct` key list order?

lib/axon/layers.ex
```diff
@@ -2524,10 +2524,10 @@ defmodule Axon.Layers do
...
 
   deftransformp stack_columns_transform(container, ignore) do
-    container
-    |> Map.from_struct()
-    |> Enum.reject(fn {k, _} -> k in ignore end)
-    |> Enum.reduce([], fn {_, v}, acc -> [v | acc] end)
+    container.__struct__().__info__(:struct)
+    |> Enum.map(fn %{field: field} -> field end)
+    |> Enum.reject(fn k -> k in ignore end)
+    |> Enum.reduce([], fn k, acc -> [Map.fetch!(container, k) | acc] end)
```
polvalente, line 2530:
```suggestion
    |> Enum.reduce([], fn %{field: k}, acc -> 
      if k in ignore do
        acc
      else
        [Map.fetch!(container, k) | acc] 
      end
    end)
```

Issue #518, opened 2023-08-12
`Axon.map_nodes`documentation examples are wrong


robinmonjo:
First example is

```
## Examples

  One common use of this function is to implement common
  instrumentation between layers without needing to build
  a new explicitly instrumented version of a model. For example,
  you can use this function to visualize intermediate activations
  of all convolutional layers in a model:

      instrumented_model = Axon.map_nodes(model, fn
        %Axon{op: :conv} = graph ->
          Axon.attach_hook(graph, &visualize_activations/1)

        graph ->
          graph
      end)
```

However, `map_nodes` will pass an `Axon.Node` and not an `Axon` so this example doesn't compile and we can't use `Axon.attach_hook` on an `Axon.Node`.

I haven't tested the second example but it seems wrong as well.

What should we do about this ? We can fix the first example but that seems weird that we don't have an API to attach a hook to a node directly without doing things like:

```
instrumented_model = Axon.map_nodes(model, fn
  %Axon.Node{op: :conv, id: id} = node ->
    fun = fn container ->
      IO.inspect(container, label: id)
    end
    %{node | hooks: [{:forward, :both, fun} | node.hooks]}
  node ->
    node
end)
```

seanmor5:
Good catch, I think attach_hook should probably be modified to work on nodes directly as well as the parent model. Potentially we can also introduce node manipulation APIs in `Axon.Node`, but I would like to avoid that for now.

PR #519, opened 2023-08-12
make attach_hook API compatible with Axon.Node and fix doc example on Axon.map_nodes

robinmonjo:
Fix: https://github.com/elixir-nx/axon/issues/518


seanmor5 (commented):
Thank you!

PR #520, opened 2023-08-15
test: ensure Loop.checkpoint reacts to event filter

polvalente:


test/axon/loop_test.exs
```diff
@@ -787,6 +787,29 @@ defmodule Axon.LoopTest do
...
+               |> Loop.checkpoint(event: :iteration_completed, filter: [every: 2])
+               |> Loop.run(data, %{}, epochs: 3)
+
+      assert [
+               "checkpoint_0_0.ckpt",
+               "checkpoint_0_2.ckpt",
+               "checkpoint_0_4.ckpt",
+               "checkpoint_1_1.ckpt",
+               "checkpoint_1_3.ckpt",
+               "checkpoint_2_0.ckpt",
```
polvalente, line 806: @seanmor5 looks like checkpointing does work (probably I did something wrong on my end). Though I didn't expect the event count (and thus the filter) for iterations to be a total tally over the whole run. Is this expected?

polvalente: Right. I can see the argument for both. Just wanted to make sure this is the expected format.

PR #524, opened 2023-08-21
Add block utility

### LINKED ISSUE #169: Add weight sharing

### LINKED ISSUE #522: Trouble implementing models in Axon
Recently I tried coding a somewhat popular model in Axon, only to give up after few hours. The reason is that I found implementing custom models very hard, if not impossible. It could be that I am missing something, that I need more practice, or that I am biased towards PyTorch that I was using for years. Still, I would like to ask some questions.
Is there any particular reason why `Axon.layer_name` was chosen to return a function and not a struct with parameters? I know the latter is more OOP than functional (as stated [here](https://github.com/elixir-nx/axon/issues/475#issuecomment-1427086452)), but implementing custom models to me seems simpler this way and it allows for parameters to be reused easily. Also, I think this is exactly the approach that was taken in `Nx.Scholar`.
Many popular algorithms involve applying the same set of weights over and over again to the input. For example, recurrent neural networks (potentially with custom cells), [Deep Sets](https://arxiv.org/abs/1703.06114), meta-learning ([MAML](https://arxiv.org/abs/1703.03400) and related algorithms), [Neural Cellular Automata](https://distill.pub/2020/growing-ca/), etc. Currently, I don't think it is even possible to implement these in `Axon` (not sure what is going on with `get_parameters` and `set_parameters` functions).
Am I missing something? Or are there any plans to change the approach? I am aware of [Add weight sharing issue](https://github.com/elixir-nx/axon/issues/169) still being open.

seanmor5:
Resolves #169 
Resolves #522 

This needs way more tests and a look at the efficiency of compiling models with block. We need to make sure using them is deterministic and that using them does not explode the graph


krstopro:
@seanmor5 Sorry, took me a while to have a look.
This seems amazing, just as I would like it to be. Elegant, yet completely functional.
I am still coding a model with it, might post an update later.

PR #535, opened 2023-10-09
Add metadata to layers

seanmor5:
This attaches Axon specific metadata to layers in the form of the `op_name`. This allows backends and compilers to match specifically on these patterns if they're available and provide optimized versions of layers rather than depending on the lower-level Nx implementation.

### DIFF
--- a/lib/axon/compiler.ex
+++ b/lib/axon/compiler.ex
@@ -933,7 +933,8 @@ defmodule Axon.Compiler do
           out
 
         %Axon.StatefulOutput{output: out} = stateful ->
-          %{stateful | output: Nx.Defn.Expr.metadata(Nx.Defn.Expr.tensor(out), %{axon_layer: op_name})}
+          out = Nx.Defn.Expr.metadata(Nx.Defn.Expr.tensor(out), %{axon_layer: op_name})
+          %{stateful | output: out}
 
         out ->
           Nx.Defn.Expr.metadata(Nx.Defn.Expr.tensor(out), %{axon_layer: op_name})
--- a/lib/axon/compiler.ex
+++ b/lib/axon/compiler.ex
@@ -709,6 +709,7 @@ defmodule Axon.Compiler do
         &5,
         &6,
         op,
+        op_name,
         parent_ids,
         name,
         args,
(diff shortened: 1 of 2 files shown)


lib/axon/compiler.ex
```diff
@@ -912,14 +914,29 @@ defmodule Axon.Compiler do
...
+
+      case result do
+        out when is_tuple(out) ->
+          out
+
+        %Axon.None{} = out ->
+          out
+
+        %Axon.StatefulOutput{output: out} = stateful ->
+          %{stateful | output: Nx.Defn.Expr.metadata(out, %{axon_layer: op_name})}
```
josevalim, line 936:
```suggestion
          %{stateful | output: Nx.Defn.Expr.metadata(Nx.Defn.Expr.tensor(out), %{axon_layer: op_name})}
```

lib/axon/compiler.ex
```diff
@@ -912,14 +914,29 @@ defmodule Axon.Compiler do
...
 
-        op when is_atom(op) ->
-          apply(Axon.Layers, op, args)
+        out ->
+          Nx.Defn.Expr.metadata(out, %{axon_layer: op_name})
```
josevalim, line 939:
```suggestion
          Nx.Defn.Expr.metadata(Nx.Defn.Expr.tensor(out), %{axon_layer: op_name})
```

lib/axon/compiler.ex
```diff
@@ -912,14 +914,29 @@ defmodule Axon.Compiler do
...
+
+      case result do
+        out when is_tuple(out) ->
+          out
+
+        %Axon.None{} = out ->
+          out
+
+        %Axon.StatefulOutput{output: out} = stateful ->
+          %{stateful | output: Nx.Defn.Expr.metadata(Nx.Defn.Expr.tensor(out), %{axon_layer: op_name})}
```
josevalim, line 936:
```suggestion
          out = Nx.Defn.Expr.metadata(Nx.Defn.Expr.tensor(out), %{axon_layer: op_name})
          %{stateful | output: out}
```

josevalim:
@seanmor5 it should be good now :)

josevalim:
FWIW, the failing tests should be fixed by using Nx.all_close or similar. :)

PR #536, opened 2023-10-18
Provide layer name as hook name

seanmor5:
This allows us to override the hook if necessary. I was searching for a solution to provide Layer metadata in the hook function signature, but I don't think it's possible without some serious work arounds


josevalim (commented):
Do we need docs and/or tests? :)

PR #542, opened 2023-11-09
Update horses_or_humans.livemd with the correct dataset url

weij:
Run into the following error with the original url. It seems that the dataset has been moved.
```
<Error>
<Code>NoSuchBucket</Code>
<Message>The specified bucket does not exist.</Message>
</Error>
```


seanmor5:
Fixed in #543

PR #553, opened 2023-12-30
Use model state struct instead of parameters

seanmor5:
cc @jonatanklosko @josevalim 

Almost all tests passing, have to rebase as well

### DIFF
--- a/lib/axon/compiler.ex
+++ b/lib/axon/compiler.ex
@@ -172,12 +172,16 @@ defmodule Axon.Compiler do
           params
         end)
 
-      model_state_meta
-      |> normalize_blocks(params)
+      out =
+        params
+        |> normalize_blocks(model_state_meta)
+        |> merge_model_state!(init_model_state)
 
       if debug? do
         Logger.debug("Axon finished init expression generation in #{us_to_ms(time)}ms")
       end
+
+      out
     end
 
     {init_fun, predict_fun}
@@ -285,6 +289,10 @@ defmodule Axon.Compiler do
     end
   end
 
+  defp merge_model_state!(state, init_state) do
+    %{state | data: merge_params!(state.data, init_state.data)}
+  end
+
   defp merge_params!(params, init_params) do
     Enum.reduce(init_params, params, fn {key, value}, params ->
       case params do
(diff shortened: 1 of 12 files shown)


lib/axon/compiler.ex
```diff
@@ -115,7 +134,17 @@ defmodule Axon.Compiler do
 
     init_cache = Map.new(cache, fn {_, {int_id, funs}} -> {int_id, funs} end)
 
-    init_fun = fn template, init_params ->
+    init_fun = fn template, init_state ->
+      # TODO: Check if init_state and output state match
+      {init_params, meta} =
+        case init_state do
+          %Axon.ModelState{data: init_params} = model_state ->
+            {init_params, Map.take(model_state, [:parameters, :frozen_parameters, :state])}
```
josevalim, line 142: If feels a bit backwards to tear the ModelState apart and then set it up again. What if you return `model_state` here and an empty model state in the branch below, and then always update it directly instead?

If you really want to break it apart, then I would break it apart into four distinct parameters:

```suggestion
            {init_params, model_state.parameters, model_state.frozen_parameters, model_state.state}
```

lib/axon/compiler.ex
```diff
@@ -225,6 +265,30 @@ defmodule Axon.Compiler do
...
+          updated_state =
+            if model_state.state == %{}, do: state, else: Map.put(state, key, model_state.state)
+
+          {Map.put(acc, key, model_state.data), updated_parmeters, updated_state}
+
+        {key, data}, {acc, parameters, state} ->
+          {Map.put(acc, key, data), parameters, state}
+      end)
+
+    {data, %{parameters: parameters, state: state}}
```
josevalim, line 289: So what are blocks? Are those ModelState inside ModelStates? Are they the primary motivation for adding ModelState?

josevalim: I would not add a map here:

```suggestion
    {data, parameters, state}
```

Otherwise we are putting it in a map only to access individual members later.

lib/axon/compiler.ex
```diff
@@ -129,13 +158,24 @@ defmodule Axon.Compiler do
           params
         end)
 
+      {params, %{parameters: block_parameters, state: block_state}} = normalize_blocks(params)
       params = merge_params!(params, init_params)
 
+      model_state_meta =
+        model_state_meta
+        |> Map.update!(:parameters, &Map.merge(&1, block_parameters))
+        |> Map.update!(:state, &Map.merge(&1, block_state))
```
josevalim, line 167: Instead of merging after the fact, pass `model_state_meta.parameters` and `model_state_meta.state` to `normalize_blocks` and already store new key-value directly into them, avoiding this merge operation here.

Quick question: do you need to use the logic in `merge_params!` to also merge the state parameters? What is the difference between the model state data/parameters/frozen_parameters/state fields?

Especially because only the first one is kept as a container, everything else would become part of the Nx.Defn cache key, so picking one over the other matters a lot!

lib/axon/model_state.ex
```diff
@@ -0,0 +1,213 @@
...
+  Freezes parameters and state in the given model state
+  using the given mask.
+
+  The mask is an arity 1 function which takes the access path to the
+  leaf parameter and returns `true` if the parameter should be frozen
+  or `false` otherwise. With this, you can construct flexible masking
+  policies:
+
+      fn
+        ["dense_" <> n, "kernel"] -> n < 3
```
jonatanklosko, line 52:
```suggestion
        ["dense_" <> n, "kernel"] -> n < String.to_integer(3)
```

lib/axon/model_state.ex
```diff
@@ -0,0 +1,213 @@
...
+  defp nested_put(acc, [key], value) do
+    Map.update(acc, key, [value], fn values -> [value | values] end)
+  end
+
+  defp nested_put(acc, [key | rest], value) do
+    inner = nested_put(%{}, rest, value)
+    Map.update(acc, key, inner, &nested_put(&1, rest, value))
+  end
+
+  def tree_get(data, access) when is_list(access) do
```
jonatanklosko, line 186:
```suggestion
  defp tree_get(data, access) when is_list(access) do
```

lib/axon/model_state.ex
```diff
@@ -0,0 +1,213 @@
...
+    Enum.reduce(access, %{}, &Map.put(&2, &1, Map.fetch!(data, &1)))
+  end
+
+  def tree_get(data, access) when is_map(access) do
```
jonatanklosko, line 190:
```suggestion
  defp tree_get(data, access) when is_map(access) do
```

PR #554, opened 2024-02-16
Fix resize methods with kernel

jonatanklosko:
Currently the resize methods other than nearest give wrong results.


lib/axon/layers.ex
```diff
@@ -2049,7 +2049,9 @@ defmodule Axon.Layers do
         kernel_scale = Nx.max(1, inv_scale)
 
         sample_f =
-          Nx.add(Nx.iota({1, output_size}), 0.5) |> Nx.multiply(Nx.subtract(inv_scale, 0.5))
+          Nx.add(Nx.iota({1, output_size}), 0.5)
```
seanmor5, line 2052: I love PEMDAS

seanmor5 (approved):

lib/axon/layers.ex
jonatanklosko: Haha, I've just pushed a version we use in NxImage with more stuff in defn for clarity :)

PR #556, opened 2024-02-20
Fix event count for multiple handlers

hwuethrich:
If multiple event handlers are registered for the same event, the counter is incremented before each handler. This results in filters like `every: 10` to never match any events.

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -1881,14 +1881,14 @@ defmodule Axon.Loop do
   # attached to the loop.
   # TODO(seanmor5): Custom events
   defp fire_event(event, handler_fns, state, debug?) do
+    state = update_counts(state, event)
+
     handler_fns[event]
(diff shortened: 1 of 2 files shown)


seanmor5:
Good catch, thanks!

josevalim:
Could we have a test somehow? 🤔

hwuethrich:
@josevalim I'm currently testing it by always adding a second handler in the dummy loop, which causes the existing test to fail without the fix. 

Maybe not the most elegant/obvious solution. I'm happy to refactor the test if you point me in the right direction (I'm very new to axon) 😅

hwuethrich:
I've added a test for the event counting. Let me know if this is OK or if you would like me to further refactor the test.

hwuethrich:
Actually, the test now passes even without the fix because I removed the dummy handlers. Let me re-add them again...

seanmor5 (approved):

PR #557, opened 2024-02-22
fix broken references to Polaris.Optimizers in lstm_generation.livemd

benswift:
Even after adding the `:polaris` dep at the top, the code cell referencing `Polaris.Optimizers.adam/1` errors out (with an error about a missing `Nx.pow`, which seems to now be `Nx.power`.

This PR replaces all references to `Polaris.Optimizers` in the LSTM generation notebook with just `:adam`. There are seemingly other doco livebooks where similar changes need to be made, but I haven't got around to testing them yet.

### DIFF
--- a/notebooks/text/lstm_generation.livemd
+++ b/notebooks/text/lstm_generation.livemd
@@ -171,7 +171,7 @@ IO.puts("Total batches: #{Enum.count(train_batches)}")
 
 params =
   model
-  |> Axon.Loop.trainer(:categorical_cross_entropy, Polaris.Optimizers.adam(learning_rate: 0.001))
+  |> Axon.Loop.trainer(:categorical_cross_entropy, :adam)
   |> Axon.Loop.run(Stream.zip(train_batches, result_batches), %{}, epochs: 20, compiler: EXLA)
 
 :ok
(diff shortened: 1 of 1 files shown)


notebooks/text/lstm_generation.livemd
```diff
@@ -171,7 +171,7 @@ IO.puts("Total batches: #{Enum.count(train_batches)}")
 
 params =
   model
-  |> Axon.Loop.trainer(:categorical_cross_entropy, Polaris.Optimizers.adam(learning_rate: 0.001))
```
polvalente, line 174: I think the fix here is to  bump the Nx version to `~> 0.6` instead. Nx.pow was probably introduced in 0.5

polvalente: I'd bump all of axon, exla and nx

benswift: Hey @polvalente , I can confirm that bumping the deps fixes this notebook. The minimal set I needed to run the thing without error was

```elixir
[{:axon, "~> 0.6"},
 {:exla, "~> 0.7"},
 {:req, "~> 0.4"}]
```

One thing I'd be happy to do would be to go through all the `*.livemd` notebooks in the repo and perform similar updates. In fact, I've actually done that part, but found that some of the notebooks currently error out for other reasons (both in their current form, and with updated deps).

So, my question is: do you want me to get them to a state where they _all_ run without error? That might take a while - I'd have to investigate each one and figure out what's going wrong. Or are some of the notebooks now out-of-date and should be deleted/deprecated?

Thanks for the input, I just wanted to check what the overall plan for the Axon livebooks is before I do that work. I think the "livebooks as docs/guides" thing is great, btw - so I'm happy to help out where I can.


Anyway, we can probably close this issue in light of the broader fix to the livebooks in this repo.

polvalente: I think we could do a few smaller PRs instead of a single one to fix all livebooks. WDYT?

benswift: Yep, I think that's a good idea. Eat the elephant one bite at a time :)

I'll close this PR and submit some individual ones as I go through them.

PR #565, opened 2024-03-13
improved doc formatting by fixing indentation and escape characters

RicardoSantos-99:
Some sections of the documentation are displaying incorrectly in ExDoc

cc: @josevalim @jonatanklosko @seanmor5 

before:
image
image
image
image

after:
image
image
image
image

### DIFF
--- a/lib/axon/layers.ex
+++ b/lib/axon/layers.ex
@@ -1407,11 +1407,11 @@ defmodule Axon.Layers do
   ## Options
 
     * `:rate` - dropout rate. Used to determine probability a connection
-    will be dropped. Required.
+      will be dropped. Required.
 
     * `:noise_shape` - input noise shape. Shape of `mask` which can be useful
-    for broadcasting `mask` across feature channels or other dimensions.
-    Defaults to shape of input tensor.
+      for broadcasting `mask` across feature channels or other dimensions.
+      Defaults to shape of input tensor.
 
   ## References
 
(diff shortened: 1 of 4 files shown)


lib/axon/layers.ex
```diff
@@ -1407,11 +1407,11 @@ defmodule Axon.Layers do
   ## Options
 
     * `:rate` - dropout rate. Used to determine probability a connection
-      will be dropped. Required.
+    will be dropped. Required.
```
jonatanklosko, line 1410: We usually keep lists indented like that, this is not an issue :) Please correct other places

```suggestion
      will be dropped. Required.
```

lib/axon/loop.ex
```diff
@@ -186,13 +186,13 @@ defmodule Axon.Loop do
   Axon loops are typically created from one of the factory functions provided in this
   module:
 
-      * `Axon.Loop.loop/3` - Creates a loop from step function and optional initialization
-      functions and output transform functions.
+    * `Axon.Loop.loop/3` - Creates a loop from step function and optional initialization
+    functions and output transform functions.
```
jonatanklosko, line 190:
```suggestion
    * `Axon.Loop.loop/3` - Creates a loop from step function and optional initialization
      functions and output transform functions.
```

jonatanklosko (approved):
Thank you!

PR #573, opened 2024-05-11
Implement loops as Streams

### LINKED ISSUE #362: Default loop output transforms are too intrusive

### LINKED ISSUE #455: Consider making all loop executions run in a separate process
I think this would enable some pretty interesting functionality, but I haven't yet thought through any potential drawbacks. I will do a detailed proposal soon

seanmor5:
This takes the regular loop `run` and builds it based off of this `stream` which will return a loop state per epoch. Basically this should be really flexible in that I can now control the running of the loop with regular Elixir `Stream` and `Enum` functions. You can see this in how much simpler the older `run` is now. 

This also kills `output_transform` in the loop. I may add it to the `Loop.run` as an option and that case we can add some output transforms that will not break some older training code

Resolves #362

Will potentially resolve #455 depending on how my experiments go

### DIFF
--- a/lib/axon/loop.ex
+++ b/lib/axon/loop.ex
@@ -49,9 +49,7 @@ defmodule Axon.Loop do
 
       %Axon.Loop.State{
         epoch: integer(),
-        max_epoch: integer(),
         iteration: integer(),
-        max_iteration: integer(),
         metrics: map(string(), container()),
         times: map(integer(), integer()),
         step_state: container()
@@ -65,14 +63,6 @@ defmodule Axon.Loop do
   an additional initialization function to provide some starting point for the step
   state. For machine learning tasks, the initialization function will return things like
   initial model parameters and optimizer state.
-
-  Typically, the final output of the loop is the accumulated final state; however, you
-  may optionally apply an output transform to extract specific values at the end of the
-  loop. For example, `Axon.Loop.trainer/4` by default extracts trained model state:
-
-      output_transform = fn state ->
-        state.step_state[:model_state]
-      end
 
   ## Initialize and Step
 
@@ -274,7 +264,6 @@ defmodule Axon.Loop do
     :init,
     :step,
     :attached_state,
-    :output_transform,
     metrics: %{},
     handlers: @default_handlers
   ]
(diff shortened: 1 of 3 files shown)


lib/axon/loop.ex
```diff
@@ -1547,6 +1528,118 @@ defmodule Axon.Loop do
...
+        opts \\ []
+      ) do
+    {max_iterations, opts} = Keyword.pop(opts, :iterations, -1)
+    {jit_compile?, opts} = Keyword.pop(opts, :jit_compile?, true)
+    {strict?, opts} = Keyword.pop(opts, :strict?, true)
+    {force_garbage_collection?, jit_opts} = Keyword.pop(opts, :force_garbage_collection?, false)
+    debug? = Keyword.get(jit_opts, :debug, false)
+
+    sample_data =
+      case Enum.take(data, 1) do
```
josevalim, line 1558: This is tricky...

If the stream is pointing to an external resource, then you are opening it and traversing it. If the stream has side-effects, this first element is effectively lost. The correct way is to do this when streaming starts. Something like this:

```
Stream.unfold({:init, arg1, arg2}, fn
  {:init, arg1, arg2}, data ->
    state = init_loop_state(...)
    {emit_next(state), {:state, state}
  {:state, state}, data ->
    ....
  :halt, _data ->
    ...
end)
```

josevalim: I think you would have to move everything to `init`, including `fire_event(:started, ...)`.

lib/axon/loop.ex
```diff
@@ -1588,262 +1681,153 @@ defmodule Axon.Loop do
...
-    } = loop
-
-    sample_data =
-      case Enum.take(data, 1) do
-        [sample_data | _] ->
-          sample_data
+    loop
+    |> stream(data, init_state, opts)
+    |> Stream.take(max_epochs)
+    |> Stream.with_index()
```
josevalim, line 1688: Do you need `with_index` if you already have the epoch?

PR #581, opened 2024-06-10
Add inspect_values option

seanmor5:


lib/axon.ex
```diff
@@ -3761,6 +3761,10 @@ defmodule Axon do
       metrics. Also forwarded to JIT if debug mode is available
       for your chosen compiler or backend. Defaults to `false`
 
+    * `:inspect_values` - if `true`, will print intermediate layer
```
josevalim, line 3764: nitpick if we should call it `print_values` but both ways are fine!

josevalim (approved):

PR #596, opened 2024-09-22
Release v0.7.0

seanmor5:


josevalim:
@seanmor5 should we nuke part of the quantized stuff now that it has made its way into Nx?

PR #598, opened 2024-10-12
Fix: call apply/3 as intended

preciz:
Thank you for reviewing.


polvalente (approved):
I think it'd be nice to add a small test for this edge case

preciz:
I added a test that fails with the old code but succeeds with the new one.

polvalente (approved):

polvalente:
Merged with the failing EXLA test because it's already failing in main

PR #608, opened 2025-04-26
docs: add pytorch cheatsheet

polvalente:


guides/cheatsheets/axon_pytorch.cheatmd
```diff
@@ -0,0 +1,546 @@
...
+
+### Dense / Linear
+
+Applies a linear transformation to the incoming data: `y = xW^T + b`.
+
+#### Axon
+```elixir
+Axon.dense(out_features, activation: :relu, name: "my_dense_layer")
+dense_layer = Axon.dense(128)
+```
```
seanmor5, line 64: This needs the inputs! The graph always has to be built in a "connected" way:

```suggestion
```elixir
input = Axon.input("features")
Axon.dense(input, out_features, activation: :relu, name: "my_dense_layer")
dense_layer = Axon.dense(input, 128)
```
```

Comment (seanmor5) [reply guides/cheatsheets/axon_pytorch.cheatmd:64]: You can mention if you want to do the same thing as in pytorch you can use `Axon.block`. That will turn a sub-graph into a re-usable function

File: guides/cheatsheets/axon_pytorch.cheatmd [hunk 2]
```diff
@@ -0,0 +1,546 @@
...
+relu = nn.ReLU()
+output = relu(dense_layer(x))
+```
+
+### Convolutional (Conv2D)
+
+Applies a 2D convolution over an input signal composed of several input planes.
+
+#### Axon
+```elixir
```
seanmor5, line 78: Same as above

guides/cheatsheets/axon_pytorch.cheatmd
polvalente: There are few times where I want to react with a 🤦‍♂️ emoji.

This is one of those times.

guides/cheatsheets/axon_pytorch.cheatmd
```diff
@@ -0,0 +1,546 @@
...
+```
+
+## Activation Functions
+
+#### Axon
+
+Activations are typically specified as options within layers (like `Axon.dense`) or applied as separate layers in the model definition pipeline.
+
+```elixir
+# Option 1: As layer option
```
seanmor5, line 213: You can also do:

```
Axon.activation(:softmax)
```

guides/cheatsheets/axon_pytorch.cheatmd
```diff
@@ -0,0 +1,546 @@
...
+
+#### Axon
+
+Axon allows for the definition of custom layers and models.
+
+```elixir
+# Example:
+defmodule MyCustomLayer do
+  def call(x) do
+    Axon.dense(128, activation: :relu, name: "my_dense_layer")
```
seanmor5, line 251:
```suggestion
    Axon.dense(x, 128, activation: :relu, name: "my_dense_layer")
```

seanmor5: This would be a good section to mention `Axon.block`, which is kind of like an `nn.Module` in that it creates a re-usable block

PR #625, opened 2026-01-31
Correct implementation of weight tying

seanmor5:
Resolves #325 

Simple weight tying implementation via shared parameters in the model state. These are resolved to the correct parameter, and allow an optional transform (e.g. `Nx.transpose/1`) so the parameter shape matches effectively.

### DIFF
--- a/lib/axon/compiler.ex
+++ b/lib/axon/compiler.ex
@@ -916,23 +916,11 @@ defmodule Axon.Compiler do
     else
       # Parameters are just accessed in the layer sub-map of the nested
       # parameter map, so we just need to extract them and then apply
-      # freezing and dtype policy
+      # freezing and dtype policy. Parameters may be SharedParameter
+      # structs for tied weights, which are resolved to their source.
       parameter_inputs =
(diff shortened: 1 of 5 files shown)


lib/axon/compiler.ex
```diff
@@ -1188,5 +1176,42 @@ defmodule Axon.Compiler do
...
   defp propagating_none?(_), do: false
 
+  defp resolve_parameter!(params, layer_name, param_name, freeze?, policy) do
+    # Special case where this is a SharedParameter at the layer level, so we
+    # need to resolve that before forwarding. Otherwise this falls through and
+    # is handled at the next step
+    layer_params =
+      with %Axon.ModelState.SharedParameter{path: path} <- params[layer_name] do
+        get_in(params, path)
+      end
```
polvalente:
```suggestion
    layer_params = 
      case params[layer_name] do
        %Axon.ModelState.SharedParameter{path: path} ->
          get_in(params, path)
        nil ->
          nil
      end
```

I think this is slightly more readable

polvalente (approved):
Minor style comments

PR #632, opened 2026-03-12
Fix mix docs build failure

aphillipo:


polvalente (approved):
Oops, I think I didn't commit this fix. Thanks!

aphillipo:
@polvalente just noticed this mix docs wasn't working by default, let me know if this is a sensible fix?

aphillipo:
I think I may need to add my new serialisation guide to the docs list...

mix.exs
polvalente:
```suggestion
  @version "0.8.1"
```

In fact, I think this also got through the cracks. The published version has the correct number

polvalente:
@aphillipo feel free to submit a PR! I forgot about updating the docs list. I can update docs without a new release, so we can fix it right away!

PR #634, opened 2026-05-21
Add Swiglu activation

seanmor5:
This is fairly common in transformer implementations

### DIFF
--- a/lib/axon/activations.ex
+++ b/lib/axon/activations.ex
@@ -803,14 +803,20 @@ defmodule Axon.Activations do
 
   @doc ~S"""
   Swish-gated linear unit activation.
+
   SwiGLU splits the input tensor along the given axis into two equal
   halves $a$ and $b$, then returns $silu(a) \odot b$. The dimension of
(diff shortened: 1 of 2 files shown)


lib/axon/activations.ex
```diff
@@ -801,6 +801,51 @@ defmodule Axon.Activations do
...
+  ### Error cases
+      iex> Axon.Activations.swiglu(Nx.tensor([1.0, 2.0, 3.0]))
+      ** (ArgumentError) axis -1 of input to swiglu must have a dimension divisible by 2, got dimension of size 3
+  ## References
+    * [GLU Variants Improve Transformer](https://arxiv.org/abs/2002.05202)
+  """
+  defn swiglu(x, opts \\ []) do
+    opts = keyword!(opts, axis: -1)
+    {a, b} = split_halves(x, opts[:axis])
+    silu(a) * b
```
polvalente, line 839: @seanmor5 how about adding our first Nx.block here?

PR #636, opened 2026-05-25
Annotate activation functions with Nx blocks

seanmor5:
This is related to the introduction of `Nx.block` and custom calls. To make this transition easier, I added a `defblock` construct which marks a function as an `Nx.block` with a generated camel-case module name equivalent to the function name.

The plan would be to use this for all layers as well, but that requires some changes because some layers are `deftransform` instead of `defn`.

Axon blocks will be prefixed with `Axon.Block`. It's confusing a bit because these are different than `Axon.block`

### DIFF
--- a/lib/axon/activations.ex
+++ b/lib/axon/activations.ex
@@ -44,6 +44,7 @@ defmodule Axon.Activations do
   """
 
   import Nx.Defn
+  import Axon.Block
   import Axon.Shared
 
   @doc ~S"""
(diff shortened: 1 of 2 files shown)


lib/axon/activations.ex
```diff
@@ -759,7 +760,7 @@ defmodule Axon.Activations do
       >
 
   """
-  defn softplus(x) do
+  defblock Axon.Block, softplus(x) do
```
polvalente, line 763: @josevalim @seanmor5 wdyt about having a single Axon.Block struct with module, function and arity as the name?

```elixir
%Axon.Block{module: Axon.Activations, function: :softplus, arity: 1, options: ...}

%Axon.Block{module: Axon.Layers, function: :dense, arity: 2, options: ...}
```

or something like that

polvalente: Actually, I think we should have it like this:

`defblock selu` -> `Axon.Activations.Selu`
`defblock SeLU, selu` -> `Axon.Activations.SeLU`

Where the block name is given by a module suffix optionally. This makes it easier to define, avoids accidental name collisions and the optional name argument gives flexibility for non-standard casing.

polvalente:
closed in favor of #639

PR #637, opened 2026-05-27
Add a residual layer combinator

seanmor5:
Simple combinator which makes residuals a bit more natural/expressive


polvalente (approved):
Could use tests I think, but looks good otherwise

PR #639, opened 2026-08-07
Annotate activation functions with Nx blocks

Chapaman:
## Summary

- Continues [elixir-nx/axon#636](https://github.com/elixir-nx/axon/pull/636)


lib/axon/activations.ex
```diff
@@ -517,13 +517,19 @@ defmodule Axon.Activations do
...
-      Nx.max(x, 0),
+      relu_block(x),
       [x],
       fn g -> [Nx.select(Nx.greater(x, 0), g, Nx.broadcast(0, g))] end
     )
   end
 
+  defblock Relu, relu_block(x) do
+    Nx.max(x, 0)
+  end
```
polvalente: I'm not sure that custom_grad should sit outside of Nx.block. The grad of block is the grad of the default callback IIRC, so I don't think this should matter

Chapaman: > I'm not sure that custom_grad should sit outside of Nx.block. The grad of block is the grad of the default callback IIRC, so I don't think this should matter

You're right on the default callback, The thing I'm having trouble with is on forwarding. The callback gets re-run on real tensors and `custom_grad`/`stop_grad` turns results into a traced tensor

polvalente: `custom_grad` and `stop_grad` should work though. Maybe `defblock` is not using proper `defn` scoping for the callback body?

Chapaman: maybe we should make sure that `custom_grad`and `stop_grad` are guarded on Nx?

lib/axon/activations.ex
```diff
@@ -686,54 +676,43 @@ defmodule Axon.Activations do
...
-      |> reciprocal()
-      |> Nx.multiply(stable_exp)
-
-    # Cache logits so they are available in certain calculations,
-    # e.g. binary_cross_entropy and categorical_cross_entropy
-    cache_logits(x, res)
+    stable_exp
+    |> Nx.sum(axes: axes, keep_axes: true)
+    |> reciprocal()
+    |> Nx.multiply(stable_exp)
```
polvalente, line 715: cache_logits seems to be missing from this implementation

lib/axon/activations.ex
```diff
@@ -686,54 +676,43 @@ defmodule Axon.Activations do
...
       >
 
   """
   defn softmax(x, opts \\ []) do
+    # Logits metadata must wrap the block result (not sit inside it) so
+    # losses can pattern-match `%{logits: _}` on the returned tensor.
+    cache_logits(x, softmax_block(x, opts))
+  end
+
+  defblock SoftMax, softmax_block(x, opts \\ []) do
```
polvalente: Maybe this warrants `defblockp` where the inner deftransform is also defined with `deftransformp` so that this module just advertises `softmax` but not `softmax_block`

lib/axon/block.ex
```diff
@@ -0,0 +1,181 @@
...
+      end
+
+    quote do
+      defmodule unquote(struct_module) do
+        @moduledoc false
+        unquote(struct_def)
+      end
+
+      # Public transform first so a preceding @doc attaches here, not to defnp.
+      deftransform unquote(name)(unquote_splicing(args)) do
```
polvalente:
```suggestion
     unquote(kind)(unquote(name)(unquote_splicing(args))) do
```

where `kind` is either `:deftransform` or `:deftransformp` coming from `defblock` and `defblockp` respectively, I think is the correct solution to my other comment in activations

polvalente (approved):

polvalente (approved):

PR #644, opened 2026-08-15
Fix dynamic RNN unrolling gradients, generalize unrolls into Axon.Layers.scan/5

seanmor5:
## The bug

`Axon.Layers.dynamic_unroll/7` compiled to a plain Nx `while`. Nx's built-in reverse-mode rule for `while` multiplies the per-step Jacobians in **forward** time order, which is only correct when those Jacobians commute. RNN cell Jacobians do not commute, so every dynamically unrolled LSTM/GRU/ConvLSTM was training on wrong gradients — silently, and divergently from the same model unrolled statically.

Concretely, on a GRU with random weights, comparing `static_unroll` and `dynamic_unroll` gradients of `Nx.sum(ys)`:

| | before | after |
|---|---|---|
| max abs grad diff (`wir`) | 0.430 | 6.0e-8 |
| max abs grad diff (`whn`) | 1.355 | 1.2e-7 |
| max abs grad diff (`bhn`) | 0.823 | 4.8e-7 |

`:dynamic` is the default `:unroll` for the RNN layers, so this affected RNN training out of the box.

## The fix

Both unrolls are replaced by a single `jax.lax.scan`-style primitive, `Axon.Layers.scan/5`:

```elixir
{final_carry, ys} = Axon.Layers.scan(body_fn, init, xs, params, unroll: :dynamic, axis: 0)
```

- `body_fn.(carry, x_t, params) -> {new_carry, y_t}`.
- The carry, inputs, params and outputs may each be any `Nx.Container` of tensors — a single tensor, a tuple, a map, or a nested combination.
- `unroll: :static` inlines the loop into the compilation graph.
- `unroll: :dynamic` compiles to a `while` and installs a `custom_grad` that runs a **reverse**-time loop taking a per-step VJP, which is the correct reverse-mode order.

`dynamic_unroll/7` and `static_unroll/7` are kept as backwards-compatible wrappers over `scan/5` that adapt the RNN cell signature, and the LSTM/GRU/ConvLSTM layers now dispatch through it — so they pick up the fix directly with no call-site changes.

## Graph-level `Axon.scan/4`

The second commit exposes the primitive at the graph level. The body is itself an Axon subgraph, so parameterized layers used inside it are initialized once and shared across every step:

```elixir
input = Axon.input("seq", shape: {nil, 10, 32})
init = Axon.input("h0", shape: {nil, 64})

{ys, final_carry} =
  Axon.scan(input, init, fn carry, x_t ->
    new = Axon.concatenate([carry, x_t]) |> Axon.dense(64) |> Axon.tanh()
    {new, new}
  end)
```

It returns `{ys, final_carry}` to match the ordering of `Axon.lstm/4` and `Axon.gru/4` (note `Axon.Layers.scan/5` returns the swapped `{final_carry, ys}`). The carry may be a single graph or any container of graphs — e.g. an LSTM-style `{c, h}` tuple — and the compiler routes runtime carry tensors into the body by mirroring that container's traversal order.

## Tests

- `Axon.Layers` — a regression test with a deliberately non-commuting body asserting `:static` and `:dynamic` gradients agree; the same for a composite (LSTM-style) carry and composite `xs`; forward-result equality between the two unrolls.
- `Axon.Compiler` — `Axon.scan` with a single carry, with body parameters shared across steps, with a tuple carry, `:static` vs `:dynamic` equality, and gradients flowing into body parameters.

Full suite passes (830 tests).

## Notes

- The body subgraph is wired to the compiler through reserved input names (`subgraph_x_t`, `subgraph_carry_<n>`); a user input with one of those names would collide.
- `dynamic_scan` peeks the body once to allocate the output buffer, so the body must be shape-stable across steps.

### DIFF
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -2718,6 +2718,166 @@ defmodule Axon do
   end
 
   @doc """
+  Scans `body_fun` over the time axis of `input`.
+
+  This is the graph-level counterpart to `Axon.Layers.scan/5`. The
+  body is itself an Axon subgraph: at each step `body_fun.(carry, x_t)`
+  is invoked with `%Axon{}` graphs for the per-step carry and input
+  slice, and must return `{new_carry, y_t}` as `%Axon{}` graphs (or
+  containers of them). Any layers used inside the body — including
+  parameterized layers like `Axon.dense/3` — become part of the
+  scanned cell, with parameters initialized and trained once and
+  shared across every step.
+
+  `init_carry` is the initial carry. It may be a single `%Axon{}`
+  graph or an `Nx.Container` of `%Axon{}` graphs (a tuple, a map, or
+  a nested combination), and `new_carry` returned by `body_fun` must
+  mirror its structure exactly.
+
+  Returns `{ys, final_carry}`, where `ys` stacks the per-step outputs
+  of `y_t` along `:axis` and `final_carry` is the carry after the
+  last step. The sequence-first ordering matches `Axon.lstm/4` and
+  `Axon.gru/4`; note that `Axon.Layers.scan/5` returns the swapped
+  `{final_carry, ys}` order.
+
+  ## Examples
+
+      input = Axon.input("seq", shape: {nil, 10, 32})
+      init = Axon.input("h0", shape: {nil, 64})
+
+      {ys, final_carry} =
+        Axon.scan(input, init, fn carry, x_t ->
+          new = Axon.concatenate([carry, x_t]) |> Axon.dense(64) |> Axon.tanh()
+          {new, new}
+        end)
+
+  ## Options
+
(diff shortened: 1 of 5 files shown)


lib/axon.ex
```diff
@@ -2717,6 +2717,166 @@ defmodule Axon do
...
+    * `:axis` - the time axis of `input` to scan over. Default `1`.
+
+    * `:unroll` - either `:dynamic` or `:static`. See
+      `Axon.Layers.scan/5`. Default `:dynamic`.
+
+    * `:name` - layer name.
+
+  """
+  @doc type: :recurrent
+  def scan(%Axon{} = input, init_carry, body_fun, opts \\ [])
```
polvalente, line 2765: We should have scan as an Nx function I think

PR #647, opened 2026-08-15
Support container outputs in eval_step and Axon.Display

### LINKED ISSUE #612: Axon.Loop.validate doesn't work with models that return a container
Axon.container let's us return a map of multiple outputs. For example most of the models in Bumblebee do this. `%{logits: ..., encoder_hidden_state: ...}`, but I have found that this can cause issues in quite a few places. One is `Axon.Display.as_table` which doesn't seem to handle maps/containers or Axon.None. And the other is `Axon.Loop.validate`. Even though this will call the same loss function passed into `Axon.Loop.trainer`, in the `Axon.Loop.eval_step` function it takes the model output and passes it directly into `Nx.type/1` and `Nx.shape/1` which both don't handle maps/containers. Because of this, the outputs of the model never reach the loss function. 

https://github.com/elixir-nx/axon/blob/35004b489e4a0fd5367445c57d0ce6c96f879073/lib/axon/loop.ex#L481C1-L482C38

I'm a little unclear here why this init function can't just return the model output directly. It seems to try to recreate the shape of the models output with all zeros, but only handles outputs that are a tensor. Forgive my ignorance if there's something I'm missing, but can't this just be reduced from:

```
{inp, tar}, state ->
        # TODO: Is this expensive
        output = forward_model_fn.(state, inp)
        output_type = Nx.type(output)
        output_shape = Nx.shape(output)
        y_pred = Nx.broadcast(Nx.tensor(0, type: output_type), output_shape)

        %{
          model_state: state,
          y_true: zeros_like(tar),
          y_pred: y_pred
        }
```

to:

```
{inp, tar}, state ->
        # TODO: Is this expensive
        output = forward_model_fn.(state, inp)

        %{
          model_state: state,
          y_true: zeros_like(tar),
          y_pred: output
        }
```

I'm happy to write up a PR if this idea seems reasonable, but admittedly my knowledge of the codebase and the repercussions of this change is closet to zero.

seanmor5:
Some models give back a group of tensors instead of just one, like `%{logits: ..., hidden_state: ...}`. The evaluation loop and the display functions only knew how to look at a single tensor, so they crashed on those models.

Now they walk through the whole group and handle each tensor inside it, including missing optional outputs. Closes #612.

### DIFF
--- a/lib/axon/display.ex
+++ b/lib/axon/display.ex
@@ -237,8 +237,7 @@ defmodule Axon.Display do
     rendered =
       shapes
       |> Enum.sort()
-      |> Enum.map(&render_output_entry/1)
-      |> Enum.join(", ")
+      |> Enum.map_join(", ", &render_output_entry/1)
 
     "%{#{rendered}}"
   end
--- a/lib/axon/display.ex
+++ b/lib/axon/display.ex
@@ -224,12 +224,30 @@ defmodule Axon.Display do
     "#{type}#{shape}"
   end
 
+  defp render_output_shape(%Axon.None{}), do: "none"
+
   defp render_output_shape(shapes) when is_tuple(shapes) do
     shapes
     |> Tuple.to_list()
     |> Enum.map(&render_output_shape(&1))
     |> Enum.join(", ")
   end
+
+  defp render_output_shape(%{} = shapes) do
+    rendered =
+      shapes
+      |> Enum.sort()
+      |> Enum.map(&render_output_entry/1)
+      |> Enum.join(", ")
+
+    "%{#{rendered}}"
+  end
+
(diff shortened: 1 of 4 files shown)


lib/axon/display.ex
```diff
@@ -224,13 +224,31 @@ defmodule Axon.Display do
...
     |> Enum.map(&render_output_shape(&1))
     |> Enum.join(", ")
   end
 
+  defp render_output_shape(%{} = shapes) do
+    rendered =
+      shapes
+      |> Enum.sort()
+      |> Enum.map(&render_output_entry/1)
+      |> Enum.join(", ")
```
seanmor5:
```suggestion
      |> Enum.map_join(", ", &render_output_entry/1)
```

PR #650, opened 2026-08-15
Add scaled dot-product attention

seanmor5:


lib/axon/layers.ex
```diff
@@ -2696,4 +2689,308 @@ defmodule Axon.Layers do
...
+    scores = Nx.dot(query, [3], [0, 1], key, [3], [0, 1]) * scale
+    scores = apply_attention_mask(scores, opts[:mask], mask_tensor)
+
+    weights = Axon.Activations.softmax(scores, axis: -1)
+
+    {weights, new_dropout_key} =
+      maybe_attention_dropout(weights, dropout_key, opts[:dropout_rate], opts[:mode])
+
+    output = Nx.dot(weights, [3], [0, 1], value, [2], [0, 1])
+    output = from_bhsd(output, opts[:axes])
```
polvalente, line 2808: Can't we avoid transposes by using the correct reduce axes in `dot` directly?

PR #664, opened 2026-08-23
Add axis names to Axon.input and an Axon.rename layer

### LINKED ISSUE #57: Integrate named tensors
On the high-level API, we can integrate named tensors by specifying the expected names on input:

```elixir
Axon.input(batch: nil, channels: 3, height: 224, width: 224)
```
Then we'll need to consider how these are transformed through the network. For other layers, we could consider adding an option to specify output features as keywords:

```elixir
Axon.input(batch: nil, pixels: 784)
|> Axon.dense(features: 128, activation: :relu)
|> Axon.dense(label: 10, activation: :softmax)
```

seanmor5:
Closes #57.

The issue proposed writing input shapes as keyword lists (`Axon.input(batch: nil, channels: 3, ...)`) so that axes can be referred to by name. The maintainer feedback was that mixing keyword lists and tuples is undesirable and that a separate `:names` option, as in Nx, would be preferable. Nx itself never adopted keyword shapes, and its convention is `names: [...]` on `Nx.tensor/2`, `Nx.template/3` and `Nx.rename/2`, so this PR follows that convention.

## Design

Axon layers are plain Nx ops, so names on an input tensor already flow through the graph today: `dense` keeps the batch name, `layer_norm`, `dropout`, `conv`, pooling and element-wise layers preserve names, and compiled predict functions return named tensors. What was missing was a way to declare names on the model itself, rather than relying on the caller to hand in a named tensor, and a way to name axes produced by layers. This PR adds both.

* `Axon.input(name, names: [...])` declares axis names on an input. Names are validated at graph construction time against `:shape` when one is given (Nx's own rank-mismatch error is raised), and are stored in the node opts alongside `:shape` and `:optional`. When the model runs, the compiler applies them with `Nx.rename/2` to both the predict value and the init template, so every subsequent layer, including custom `Axon.layer` closures, can use `axes: [:features]`, `Nx.axis_index(x, :seq)` and friends. If the incoming tensor already carries a different non-nil name for an axis, an `ArgumentError` is raised; matching names pass through, and where the input declares `nil` for an axis the incoming name is kept (so `nil` behaves like the `nil` dimensions in `:shape`). `:names` is only supported for inputs with a tensor shape, not container shapes.

* `Axon.rename(x, names)` is a new shape layer mirroring `Nx.rename/2`, for naming axes that a layer produced (for example the new feature axis after `Axon.dense`). This serves the issue's "output features as keywords" idea without adding a `:names` option to every layer function.

* `Axon.Display.as_table/2` renders names in the output shape and parameter columns when present, matching `Nx` inspect (`f32[batch: 2][features: 8]`). The input row's options column already lists every entry of the node opts, so `names: [...]` shows up there as well.

Usage:

```elixir
model =
  Axon.input("features", shape: {nil, 784}, names: [:batch, :pixels])
  |> Axon.dense(128)
  |> Axon.rename([:batch, :hidden])

# custom layers can refer to axes by name
input = Axon.input("features", shape: {nil, 784}, names: [:batch, :pixels])
Axon.layer(fn x, _opts -> Nx.sum(x, axes: [:pixels]) end, [input])
```

`:names` is only stored in the input node opts when given, so unnamed inputs keep exactly the same node opts (and `Axon.Display` output) as before. The compiler's input clause previously pattern-matched the exact keyword list `[shape: _, optional: _]`; it now reads the opts with `Keyword.fetch!`/`Keyword.get`, so both named and unnamed inputs compile. A small `Axon.Shape.tensor_shape?/1` helper is exposed so `input/2` can tell tensor shapes from container shapes the same way `Axon.Shape.input/1` does.

## Limitations

* Keyword-list shapes are intentionally not supported, per the maintainer feedback on the issue.
* No per-layer `:names` options are added (e.g. `Axon.dense(128, names: ...)`); `Axon.rename/3` is the escape hatch.
* `Axon.flatten` and the RNN layers still drop names, following `Nx.reshape` and the RNN loop semantics. `Axon.rename/3` can restore them after those layers.
* `:shape` is still not validated at run time (pre-existing `TODO` in the compiler); only `:names` is checked against the actual tensor.
* `Axon.get_inputs/1` is unchanged and still returns `name => shape`, since changing its return type would be breaking.

## Tests

* `test/axon_test.exs`: input nodes carry `:names`, names default to `nil`, rank mismatch against `:shape`, non-atom names, names on container shapes, names without a shape, and `Axon.rename/3` node construction and validation.
* `test/axon/compiler_test.exs`: predict renames an unnamed tensor and preserves values; names flow into the init template and `Axon.get_output_shape/2`; names are usable from custom layers via `axes: [:features]`; matching and partial names on incoming tensors pass through, and incoming names are kept where the input declares `nil`; conflicting names, rank mismatches and container values raise; omitted optional inputs with names still resolve to `%Axon.None{}`; multi-input models only rename the inputs that declare names; `Axon.rename/3` forward pass, downstream use and rank errors.
* `test/axon/display_test.exs`: named and unnamed templates render as `f32[batch: 2][features: 8]` and `f32[2][8]`, and `names: [...]` appears in the input row's options.
* `test/axon/shape_test.exs` picks up the `Axon.Shape.tensor_shape?/1` doctests.

Verified with `mix test` (905 passed, 47 excluded) and `USE_EXLA=1 mix test test/axon/compiler_test.exs test/axon/display_test.exs test/axon_test.exs` (366 passed) to cover the jitted path.

### DIFF
--- a/deps
+++ b/deps
@@ -1,0 +1,1 @@
+/Users/sean/projects/axon/deps
--- a/lib/axon.ex
+++ b/lib/axon.ex
@@ -19,6 +19,10 @@ defmodule Axon do
 
   Notice you can specify some dimensions as `nil`, indicating
   that the dimension size will be filled in at model runtime.
+  Inputs may also declare axis names with the `:names` option
+  (as in `Nx.tensor/2`), which are applied to the input tensor
+  when the model runs and propagate through subsequent layers.
+  Axes produced by layers can be named with `Axon.rename/3`.
   You can then compose inputs with other layers:
 
       model =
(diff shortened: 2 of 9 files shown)


deps
```diff
@@ -0,0 +1 @@
+/Users/sean/projects/axon/deps
```
polvalente, line 1: stray file

PR #666, opened 2026-08-23
Add Axon.Pruning with magnitude-based unstructured pruning

### LINKED ISSUE #120: Add model pruning

seanmor5:
Closes #120.

Pruning is a standard compression step alongside quantization, but Axon had nothing for it: there was no way to zero out low-magnitude weights, no way to keep them at zero while fine-tuning, and no way to report how sparse a model state is. This PR adds `Axon.Pruning`, a sibling of `Axon.Quantization` that works purely on an `Axon.ModelState` and leaves the model graph untouched.

## Design

This is a first, deliberately narrow increment: **unstructured magnitude pruning**. The parameters with the smallest absolute values are set to zero, either against one global threshold (`scope: :global`, the default, so some tensors end up sparser than others) or per tensor (`scope: :tensor`, every selected tensor is pruned to exactly the requested sparsity). Pruning always produces a *mask* alongside the pruned state: a nested map mirroring the model state data with a `{:u, 8}` tensor (`1` = keep, `0` = pruned) for every pruned parameter. Parameters that were not selected are simply absent from the mask.

The public API is:

* `magnitude_mask/3` computes the mask without touching the state. Candidates come from the `parameters` tree only, so batch-norm running statistics (which live in `state`) are never pruned. Tied parameters (`Axon.ModelState.SharedParameter`), quantized parameters (`Axon.Quantization.QTensor`) and non-float tensors are skipped. The `:filter` option receives the same access paths as `Axon.ModelState.freeze/2` (e.g. `["dense_0", "kernel"]` or `["lstm_0", "input_kernel", "wii"]`); the default keeps any path with a name ending in `"kernel"`, which covers dense/conv/embedding kernels and the composite RNN kernels while leaving biases and normalization parameters alone.
* `apply_mask/2` zeroes the masked entries of a model state or a plain parameter map using `Nx.select/3`, so dtypes are preserved and `NaN`/`Inf` don't leak through a multiply.
* `prune/3` is `magnitude_mask/3` + `apply_mask/2` and returns `{pruned_state, mask}`.
* `masked_optimizer/2` wraps anything `Axon.Loop.trainer/4` accepts (a `Polaris.Optimizers` atom or an `{init_fn, update_fn}` tuple) via `Polaris.Updates.stateful/3`. The mask is stored in the optimizer state, and the *final* update is zeroed at pruned positions, so a weight that starts at zero stays exactly zero regardless of momentum or weight decay. This composes with the existing loop and needs no changes to `Axon.Loop`. Mask entries for parameters that are not trained (e.g. frozen ones) are ignored.
* `sparsity/1` and `global_sparsity/1` report the fraction of exactly-zero entries per parameter (nested maps for composite parameters) and size-weighted across the whole state.

The mask computation ranks magnitudes with a double `Nx.argsort` (stable), so exactly `round(sparsity * n)` entries are pruned even when magnitudes tie, and it works inside `defn` on any backend. `magnitude_mask/3`, `apply_mask/2` and `prune/3` are `deftransform`s so they can be called from `defn`; the sparsity reporting functions call `Nx.to_number/1` and are eager only.

## Usage

```elixir
{pruned_state, mask} = Axon.Pruning.prune(model_state, 0.9)

Axon.Pruning.global_sparsity(pruned_state)
#=> 0.87...

trained_state =
  model
  |> Axon.Loop.trainer(:categorical_cross_entropy, Axon.Pruning.masked_optimizer(:adam, mask))
  |> Axon.Loop.run(data, pruned_state, epochs: 2)
```

## Limitations

* Zeroed weights are still stored densely: this reduces the number of non-zero parameters, not memory or latency. Getting a speedup needs sparse tensor support or structured pruning, neither of which exists yet.
* Structured/channel pruning, gradual pruning schedules and rewriting the model graph are out of scope for this increment.
* The default filter is a naming heuristic on `"kernel"`; anything else needs an explicit `:filter`.
* `masked_optimizer/2` copies the mask to the binary backend so it can be embedded in the traced optimizer init; from then on it travels as ordinary optimizer state (and is therefore checkpointed with it).
* Ties between equal magnitudes are broken by position, deterministically but arbitrarily.

## Tests

`test/axon/pruning_test.exs` covers per-tensor and global masks against hand-computed expectations, exact counts under ties, the sparsity `0`/`1` short-circuits, the default filter on dense + batch norm (state untouched) and LSTM (composite kernels masked, biases not), custom filters, skipping of tied, quantized and integer parameters, argument validation, `apply_mask/2` on model states and plain parameter maps (ignoring mask keys that are absent), all three transforms inside `defn`, `sparsity/1`/`global_sparsity/1` including composite parameters and empty states, and `masked_optimizer/2` with both `:adam` and an `{init_fn, update_fn}` tuple through a real `Axon.Loop` run (pruned entries stay exactly zero, unmasked entries train, and a control run without the wrapper does *not* keep them at zero). The suite passes on the default backend and with `USE_EXLA=1`.

### DIFF
--- a/lib/axon/pruning.ex
+++ b/lib/axon/pruning.ex
@@ -287,7 +287,7 @@ defmodule Axon.Pruning do
 
   defp optimizer_fns(optimizer) when is_atom(optimizer) do
     if Code.ensure_loaded?(Polaris.Optimizers) and
-         function_exported?(Polaris.Optimizers, optimizer, 0) do
+         {optimizer, 0} in Polaris.Optimizers.__info__(:functions) do
       apply(Polaris.Optimizers, optimizer, [])
     else
       invalid_optimizer!(optimizer)
--- a/lib/axon/pruning.ex
+++ b/lib/axon/pruning.ex
@@ -1,0 +1,422 @@
+defmodule Axon.Pruning do
+  @moduledoc """
+  Model pruning.
+
+  Pruning removes weights from a trained model by setting them to zero.
+  Most of the parameters in a trained network contribute very little to
+  its output, so a large fraction of them can be removed with a small
+  loss in accuracy, especially when the pruned model is fine-tuned for
+  a few more steps afterwards.
+
+  This module implements **unstructured magnitude pruning** of an
+  `Axon.ModelState`: the parameters with the smallest absolute values
+  are zeroed, either against a single global threshold or per tensor.
+  Pruning produces a *mask* alongside the pruned state. The mask is a
+  nested map that mirrors the model state data, with a `{:u, 8}` tensor
+  (`1` = keep, `0` = pruned) for every pruned parameter. Use it with
+  `masked_optimizer/2` to keep pruned weights at zero while fine-tuning:
+
+      # Prune 90% of the kernel weights
+      {pruned_state, mask} = Axon.Pruning.prune(model_state, 0.9)
+
+      # Report the resulting sparsity
+      Axon.Pruning.global_sparsity(pruned_state)
+      #=> 0.87...
+
+      # Fine-tune the remaining weights; pruned weights stay at zero
+      trained_state =
+        model
+        |> Axon.Loop.trainer(
+          :categorical_cross_entropy,
+          Axon.Pruning.masked_optimizer(:adam, mask)
+        )
+        |> Axon.Loop.run(data, pruned_state, epochs: 2)
+
+  ## Limitations
+
+  Zeroed weights are still stored densely; this module reduces the number
+  of non-zero parameters, not the memory footprint or the latency of the
+  model. Taking advantage of sparsity requires backend support for sparse
+  tensors or structured pruning, neither of which is implemented here. The
+  model graph is never modified, only the model state. Ties between
+  parameters of equal magnitude are broken arbitrarily but deterministically.
+  """
+  import Nx.Defn, only: [deftransform: 1, deftransform: 2]
+
+  alias Axon.ModelState
+
+  @doc """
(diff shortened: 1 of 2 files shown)


lib/axon/pruning.ex
```diff
@@ -0,0 +1,422 @@
...
+    Polaris.Updates.stateful(
+      optimizer_fns(optimizer),
+      fn _params -> %{mask: mask} end,
+      fn updates, %{mask: mask} = state, _params -> {mask_tree(updates, mask), state} end
+    )
+  end
+
+  defp optimizer_fns(optimizer) when is_atom(optimizer) do
+    if Code.ensure_loaded?(Polaris.Optimizers) and
+         {optimizer, 0} in Polaris.Optimizers.__info__(:functions) do
```
polvalente, line 290:
```suggestion
         function_exported?(Polaris.Optimizers, optimizer, 0) do
```

lib/axon/pruning.ex
```diff
@@ -0,0 +1,422 @@
...
+      |> parameter_leaves()
+      |> Enum.reduce({0, 0}, fn {_path, tensor}, {zeros, total} ->
+        {zeros + zero_count(tensor), total + Nx.size(tensor)}
+      end)
+
+    ratio(zeros, total)
+  end
+
+  defp zero_count(tensor) do
+    tensor |> Nx.equal(0) |> Nx.sum() |> Nx.to_number()
```
polvalente, line 357:
```suggestion
    tensor |> Nx.equal(0) |> Nx.as_type(:u64) |> Nx.sum() |> Nx.to_number()
```

Otherwise there will be overflows happening

PR #673, opened 2026-09-15
fix: Define Axon.ModelState.t()

georgeguimaraes:
`Axon.ModelState` has no `@type t`, so any downstream spec that says `Axon.ModelState.t()` gets an ex_doc warning about an undefined type. This adds it with the four struct fields as maps.


polvalente (commented):
Let’s not add this since we don’t validate with dialyzer. Specs like these are prone to users submitting PRs fixing the spec itself

PR #677, opened 2026-09-17
Add Axon.stop_grad/2

georgeguimaraes:
A layer that passes its input through and stops gradients from flowing back through it, the graph-level counterpart of `Nx.Defn.Kernel.stop_grad/1`. Useful when a component feeds two paths and should learn from one of them only: soothsayer's trend also scales its multiplicative seasonality and we detach it there with `Axon.nx(&Nx.Defn.Kernel.stop_grad/1)`, which this replaces with a named layer.


georgeguimaraes:
The formatting check fails on main itself (a trailing space in test/axon/loop_test.exs), #678 removes it.

